Issue #2, opened 2015-07-13
Be protocol-agnostic


swissspidy:
Although accessing a site over both HTTP and HTTPS isn't really recommended, the plugin currently doesn't work in this scenario ([see this support request](https://wordpress.org/support/topic/separate-caches-for-http-and-https?replies=2)).

The following changes need to be made:

* Take protocol into account in [`Cachify::_cache_hash`](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.class.php#L1082) method.
* Take protocol into account in [ `Cachify_HDD::_file_path`](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify_hdd.class.php#L344) method
* Add protocol to `.htaccess` rewrite rules

swissspidy:
I'd like to take a step back here.

The change proposed in #3 doesn't seem to be very backwards compatible. The old cache without the prefix wouldn't get deleted.

Also, with HTTP/2 coming up, I really recommend going HTTPS-only.

glueckpress:
@swissspidy Does this need any dev feedback? I’m afraid I can’t contribute here. :wink:

swissspidy:
@glueckpress I'll have a look at this again when I get a chance.

COLABORATI:
Hi, would you please like to add this commit to the official plugin release? I am also waiting for this, as this bug is quite annoying and I do not want to maintain that code changes manually. Thanks!

glueckpress:
@COLABORATI Wich one exactly?

swissspidy:
I'll have another look at the pull requests in February

skalscheuer:
I totally agree with you to prefer HTTPS-only pages, but this decision is not to be made by a plugin like cachify. Imho the plugin should be able to handle mixed protocols.

Code perfo is not the issue here and the drawback of doubled memory consumption is acceptable (at least not your biggest problem at this point). I'm using a modified version like PR#20 (trivial solution) and it does the job quite well.

COLABORATI:
Why no progress here? Why? Please explain. Thanks.

swissspidy:
@colaborati mainly because of lack of time and none of the pull requests being exactly what we should have in the plugin. We all do this in our spare time, so please be patient with us. If you wanna help out that would be highly appreciated

stklcode:
I just created a new PR #41 based on my previous comment. I'm using this version since approx. 5 months in production without any problems.

PR #3, opened 2015-07-14
Fix for issue #2

arnowelzel:
Fix for issue #2 as supposed in https://wordpress.org/support/topic/separate-caches-for-http-and-https but using is_ssl() instead of server variables.

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1078,8 +1078,10 @@ final class Cachify {
 
 	private static function _cache_hash($url = '')
 	{
+		if(is_ssl()) $prefix='443-';else $prefix='80-';
+		
 		return md5(
-			empty($url) ? ( $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'] ) : ( parse_url($url, PHP_URL_HOST) . parse_url($url, PHP_URL_PATH) )
+			empty($url) ? ( $prefix . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'] ) : ( $prefix . parse_url($url, PHP_URL_HOST) . parse_url($url, PHP_URL_PATH) )
 		) . '.cachify';
 	}
 
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -336,12 +336,15 @@ final class Cachify_HDD {
 
 	private static function _file_path($path = NULL)
 	{
+		if(is_ssl()) $prefix='443-';else $prefix='80-';
+		
 		$path = sprintf(
-			'%s%s%s%s',
+			'%s%s%s%s%s',
 			CACHIFY_CACHE_DIR,
 			DIRECTORY_SEPARATOR,
+			$prefix,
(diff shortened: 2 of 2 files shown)


swissspidy:
Thanks, this is now fixed in master. New release coming soon.

arnowelzel:
There was a reason why I added "443-" or "80-" as prefix and not "https-".

Look at the Apache rewrite rules:

`# GZIP FILE
<IfModule mod_mime.c>
RewriteCond %{REQUEST_URI} !^/wp/wp-admin/.*
RewriteCond %{REQUEST_METHOD} !=POST
RewriteCond %{QUERY_STRING} =""
RewriteCond %{HTTP_COOKIE} !(wp-postpass|wordpress_logged_in|comment_author)_
RewriteCond %{HTTP:Accept-Encoding} gzip
RewriteCond %{DOCUMENT_ROOT}/wp/wp-content/cache/cachify/%{SERVER_PORT}-%{HTTP_HOST}/%{REQUEST_URI}/index.html.gz -f
RewriteRule ^(.*) /wp/wp-content/cache/cachify/%{SERVER_PORT}-%{HTTP_HOST}%{REQUEST_URI}/index.html.gz [L]
 
AddType text/html .gz
AddEncoding gzip .gz
</IfModule>
 
# HTML FILE
RewriteCond %{REQUEST_URI} !^/wp/wp-admin/.*
RewriteCond %{REQUEST_METHOD} !=POST
RewriteCond %{QUERY_STRING} =""
RewriteCond %{HTTP_COOKIE} !(wp-postpass|wordpress_logged_in|comment_author)_
RewriteCond %{DOCUMENT_ROOT}/wp/wp-content/cache/cachify/%{SERVER_PORT}-%{HTTP_HOST}/%{REQUEST_URI}/index.html -f
RewriteRule ^(.*) /wp/wp-content/cache/cachify/%{SERVER_PORT}-%{HTTP_HOST}%{REQUEST_URI}/index.html [L] `

You may notice that `%{SERVER_PORT}` is part of the rule and how the full path is expanded in the end.

Now with either `https-` as prefix or no prefix at all the rule has to be doubled for both cases.

Anyway - it seems I have to live with it now.

swissspidy:
I understand your concern. It also reminds me to update the documentation.

Using `%{SERVER_PORT}` would definitely make these rules easier. However, I have a few concerns:

* The ports aren't guaranteed to be 443 or 80, e.g. when the site is behind a proxy like Varnish.
* Using 80 as the default prefix would make all previously cached files invalid. If we're changing the prefix, those invalid files should at least be properly deleted. Plus, lots of people would have to change their configuration or else we make their caching useless. 
* It's 2016 and there really is no reason not to use HTTPS _everywhere_

arnowelzel:
Invalid caches and no working cache access at all after an update: yes, that's a problem. It's indeed easier for me to change my local setup.

But who uses Cachify if there is already a Varnish setup?

And about "HTTPS everywhere": Many shared hosters still don't offer HTTPS at all and Let's encrypt is not easy to implement if you don't run your own server but only have shared hosting.

PR #20, opened 2015-10-28
fix #2 protocol aware caching

### LINKED ISSUE #2: Be protocol-agnostic

mcguffin:
Cache https output separately while keeping non-https files at the same place.

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1078,6 +1078,7 @@ final class Cachify {
 
 	private static function _cache_hash($url = '')
 	{
+		$prefix = is_ssl() ? '443-' : '';
 		return md5(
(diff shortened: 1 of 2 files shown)


swissspidy:
Thanks, this is now fixed in master. New release coming soon.

PR #22, opened 2015-11-10
Add URL scheme to hashkey.

ilyaevseev:
When Wordpress instance is accessed as http://site, Cachify stores page contents containing http:// links.

When somebody accesses https://site, then already generated page is returned, and all http:// links are blocked by web browser as cross-origins.

So cache keys should include URL scheme (http or https).

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1079,7 +1079,8 @@ final class Cachify {
 	private static function _cache_hash($url = '')
 	{
 		return md5(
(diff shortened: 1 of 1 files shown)


swissspidy:
Thanks for your PR!

Although this would solve issue #2, nowadays I don't see why people aren't going 100% HTTPS. That's why have been hesitating to merge similar pull requests.

Other PRs changed `_file_path()` as well... I guess that is needed here too.

skalscheuer:
Second issue, the protocol check is not complete. $_SERVER['HTTPS'] value can be 'off' (rarely, but possible [[doc](http://php.net/manual/en/reserved.variables.server.php)]). You should use WP's `is_ssl()` or at least add one more check.

swissspidy:
This is now fixed in master. New release coming soon.

PR #24, opened 2016-03-25
Fix #23

### LINKED ISSUE #23: Multisite, activatd by-site: clearing HDD cache clears the entire network cache
A blog admin is able to clear all other sites caches as well. 
Same may be true for Memcache, APC and DB cache.

mcguffin:
set hdd cache dir dynamically from current domain name

### DIFF
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -116,8 +116,17 @@ final class Cachify_HDD {
 
 	public static function clear_cache()
 	{
+		if ( ! is_multisite() || ! is_subdomain_install() || is_network_admin() )
+		{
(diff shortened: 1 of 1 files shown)


swissspidy:
Looks like this PR tries to fix both #23 and #2 at once. It would be great if you could only submit a PR for #23 (e.g. the last two commits).

Regarding 3c6b9e91b3a19fef720c640f7f164511625d654b, please move the `if ( ! is_multisite() || ! is_subdomain_install() || is_network_admin() )` part into the `_cache_dir()` method instead of removing that method. The method exists for a reason :-)

Zodiac1978:
@swissspidy Can this be closed because of #41?

PR #25, opened 2016-03-31
Prevent the inline css for the admin bar icon to overwrite other items

idvl:
The current inline CSS for the admin bar trash icon overwrites other admin bar items, especially if these items are using an `.ab-icon` with dashicons.

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -705,7 +705,7 @@ final class Cachify {
 		}
 
 		/* Display the admin icon anytime */
-		echo '<style>#wp-admin-bar-cachify{display:list-item !important} .ab-icon{margin:0 !important} .ab-icon:before{content:"\f182";top:2px;margin:0}</style>';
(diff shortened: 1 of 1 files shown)


swissspidy:
The change seems legit. Any specific admin bar item you've experienced this happening with?

idvl:
Yup, on the core menu items: notice the overwritten margin of all `.ab-icon` child nodes ("home" icon, "plus" icon on the "new" item). 

And if you add a custom menu item, the icon will be overwritten even if you add an icon with the `dashicons` classes:

```
add_action( 'admin_bar_menu', function ( &$wp_admin_bar ) {
    $wp_admin_bar->add_node( array(
				'id'    => 'some-id',
				'title' => '<span class="ab-icon dashicons dashicons-info"></span><span class="ab-label">My Item</span>',
				'href'  => '#',
				'meta'  => array()
			) );
} , 99 );
```

zottto:
I also found this issue when using Cachify together with the (almost unknown) WP Upstream plugin. The plugin icon now also is the trash icon instead of an "upload" icon. ;-)

PR #38, opened 2016-08-05
add <style> to wp_head() not to wp_footer()

visionlike:
FILE: cachify.class.php
```

line ->144

		/* Flush Style to wp_head() Frontend */
		add_action(
			'wp_head',
			array(
				__CLASS__,
				'add_icon_style_to_head'
			),
			90
		);
		
		/* Flush Style to wp_head() Backend */
		add_action(
			'admin_enqueue_scripts',
			array(
				__CLASS__,
				'add_icon_style_to_head'
			),
			90
		);
```

```
line ->740

	/**
	* Hinzufügen des Stylesheets für das Admin-Bar-Menü
	*
	* @since   1.2
	* @change  2.2.2
    *
    * @hook    mixed   cachify_user_can_flush_cache
	*
	* @param   object  Objekt mit Menü-Eigenschaften
	*/	
	
	function add_icon_style_to_head(){
		
		/* Aussteigen */
		if ( ! is_admin_bar_showing() OR ! apply_filters('cachify_user_can_flush_cache', current_user_can('manage_options')) ) {
			return;
		}		
		
		/* Display the admin icon anytime */
		echo  '<style>#wp-admin-bar-cachify{display:list-item !important} .ab-icon{margin:0 !important} .ab-icon:before{content:"\f182";top:2px;margin:0}</style>';
	
	}
```


swissspidy:
Hey there

It looks like you wanted to create an issue, but instead created a pull request from an old / unrelated branch. As for your suggestion, adding the stylesheet to the `<head>` instead of in the footer, I can understand where it's coming from.

According to the HTML specification `<style>` are only valid in the header, but the admin bar is loaded much later in the footer. So when hooking into `wp_head` it's not guaranteed that the admin bar is really being shown as other plugins can still disable it. Luckily, browsers usually support CSS being loaded in the footer, so it's not a big deal.

However, I've now opened #39 to see how we can improve the way we load the CSS right now.

swissspidy:
Closing in favour of #39

PR #41, opened 2016-08-13
fix #2 protocol aware caching

### LINKED ISSUE #2: Be protocol-agnostic

stklcode:
A fixed version of PR #20 by @mcguffin.

* Prefix cannot be contained in regular URLs to avoid conflicts
* Protocol prefix has effect within every caching method

Fixes #2

### DIFF
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -336,7 +336,7 @@ final class Cachify_HDD {
 
 	private static function _file_path($path = NULL)
 	{
-		$prefix = is_ssl() ? 'https$' : '';
+		$prefix = is_ssl() ? 'https-' : '';
 
 		$path = sprintf(
 			'%s%s%s%s%s',
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -336,10 +336,13 @@ final class Cachify_HDD {
 
 	private static function _file_path($path = NULL)
 	{
+		$prefix = is_ssl() ? 'https$' : '';
+
 		$path = sprintf(
-			'%s%s%s%s',
+			'%s%s%s%s%s',
 			CACHIFY_CACHE_DIR,
 			DIRECTORY_SEPARATOR,
+			$prefix,
 			parse_url(
 				'http://' .strtolower($_SERVER['HTTP_HOST']),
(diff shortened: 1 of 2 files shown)


swissspidy:
Why is the prefix `s$-` and not something that makes more sense, like `https-`?

inc/cachify_hdd.class.php
```diff
@@ -336,10 +336,13 @@ public static function _dir_size($dir = '.')
 
 	private static function _file_path($path = NULL)
 	{
+	  $prefix = is_ssl() ? 's$-' : '';
```
swissspidy: Wrong indentation

swissspidy:
Apart from the indentation and my question regarding the prefix this is 👍

stklcode:
When I wrote this, i thought about the (even really unlikely) chance that one might be running on two domains like _example.com_ and _https-example.com_ with the same post url (e.g. non '/') which would result in the same hash for different pages. That's why I've chosen $ as a delimiter.

Issue may be related to multisite bug #23.

Changed it to a more speaking `https$` though (and fixed the indentation)

swissspidy:
I don't know about the others, but I'm 100% sure there's nobody running `xy.com` _and_ `https-xy.com` _and_ running WordPress _and_ using Cachify. Therefore I prefer `https-` with a dash as a delimiter.

Would you mind changing that (in a new commit, so we don't lose history like before)? Thank you.

#23 is only somewhat related as #24 tries to fix #2 and #23. Left a comment.

stklcode:
As I said, highly unlikely ... but not impossible. 

But I'm with you, the "risk" is acceptable, made a new commit.

PR #45, opened 2016-08-21
Comment translation

stklcode:
Although if the main dev team is German, the code is available to the public and the common language should be English. Some parts already are, but mostly older comments are kept German.

Cleaned up the mixture and corrected some PHPDoc values, so that it's consistent now.
No code has been changed, shouldn't be worth an issue.


swissspidy:
Thank you very much! We'll eventually overhaul all inline documentation to align with WordPress coding standards, so this is a great start.

PR #58, opened 2017-01-02
php error fixed

timse201:


bueltge:
it is not necessary, but I think we should end each line of a statement with `;`. Like in this request on `<?php esc_html_e( 'Post/Pages-IDs (separated by commas)', 'cachify' ) ?>` to `<?php esc_html_e( 'Post/Pages-IDs (separated by commas)', 'cachify' ); ?>`.

inc/cachify.class.php
```diff
@@ -641,7 +641,7 @@ public static function add_dashboard_count( $items = array() )
 
 		/* Right now item */
 		$items[] = sprintf(
-			'<a href="%s" class="cachify-icon cachify-icon--%s" title="%s: %s">%s esc_html__( 'Cache', 'cachify' )</a>',
+			'<a href="%s" class="cachify-icon cachify-icon--%s" title="%s: %s">%s Cache</a>',
```
websupporter: Maybe we should keep it translateable?

'<a href="%s" class="cachify-icon cachify-icon--%s" title="%s: %s">%s ' . esc_html__( 'Cache', 'cachify' ) . '</a>',
`

inc/cachify.class.php
```diff
@@ -641,7 +641,7 @@ public static function add_dashboard_count( $items = array() )
 
 		/* Right now item */
 		$items[] = sprintf(
-			'<a href="%s" class="cachify-icon cachify-icon--%s" title="%s: %s">%s esc_html__( 'Cache', 'cachify' )</a>',
```
websupporter: Maybe we should keep it translateable. But I would prefer to move it to the `%s` instead of the string itself. Something like:
```
( empty($size) ? esc_html__( 'Empty Cache', 'cachify' ) : size_format($size) . ' ' . esc_html__( 'Cache', 'cachify' ) )
```

bueltge: :+1: 
Also a hint for translators would be finde, the word is to short to understand the context, like 

```php
// Translators: %s is the plugin name.
```

inc/cachify.class.php
```diff
@@ -651,7 +651,7 @@ public static function add_dashboard_count( $items = array() )
 			esc_attr(strtolower($method)),
 			esc_html__( 'Caching method', 'cachify' ),
 			esc_attr($method),
-			( empty($size) ? esc_html__( 'Empty', 'cachify' ) : size_format($size) )
+			( empty($size) ? esc_html__( 'Empty Cache', 'cachify' ) : size_format($size) . ' ' . esc_html__( 'Cache', 'cachify' ) )
```
swissspidy: This is not very translator friendly.

Should be more like

```php
/* translators: %s: cache size */
sprintf( esc_html__( '%s Cache', 'cachify' ), size_format($size) );
```

timse201: think its more friendly i did because if we let _empty_ alone it can be translated to "leer"- Cache or "leerer" - Cache the right word by word translation without context is "leer" but it should be the second one. if we let cache alone it is a noun and there are less misinterpretation. Dont know how it will be in other languages.

swissspidy: @timse201 I didn't mean the "Empty Cache" line. That's fine. Changing that would be even worse. What I mean is something along these lines:

```php
if ( empty( $size ) ) {
  esc_html__( 'Empty Cache', 'cachify' )
} else {
  /* translators: %s: cache size */
  sprintf( esc_html__( '%s Cache', 'cachify' ), size_format($size) );
}
```

Why? Because `size_format($size) . ' ' . esc_html__( 'Cache', 'cachify' ) )` is not translator friendly. Think about right-to-left languages etc.

timse201: thanks for clarify

bueltge: Again a small hint, for the last request. Please use strict comparison coding rules `$size == 0` should be `$size === 0`.

inc/cachify.class.php
```diff
@@ -639,6 +639,12 @@ public static function add_dashboard_count( $items = array() )
 			)
 		);
 
+		/* Output of the cache size */
+		$cachesize = ( $size == 0 )
+			? esc_html__( 'Empty Cache', 'cachify' ) :
+			/* translators: %s: cache size */
+			sprintf( esc_html__( '%s Cache', 'cachify' ), size_format($size) );;
```
swissspidy: Note that there's a duplicated semicolon here.

For improved readability, I would use a regular `if  ( empty( $size ) { $cachesize = … } else { $cachesize = … }` structure

timse201: think it is nicer to have $cachesize in front
but if all of you want we can do it this way and i will upload this:

'''
if ( empty( $size ) ) {
			$cachesize = esc_html__( 'Empty Cache', 'cachify' );
			}
			else {
				$cachesize =
				/* translators: %s: cache size */
				sprintf( esc_html__( '%s Cache', 'cachify' ), size_format($size) );
				}
'''

bueltge:
Last commit fix all hints of the comments, thanks again.

PR #59, opened 2017-01-03
align icons

timse201:


bueltge:
Looks good, but needs test on different clients.

timse201:
i tested it with firefox's screen size test what ive not tested is windows dpi setting

Issue #57, opened 2017-01-01
More Screenshots on WordPress.org


timse201:
at the moment there is only 1 picture on WordPress.org
https://ps.w.org/cachify/assets/screenshot-1.png?rev=976963

There sould be more screenshots
like:
https://github.com/pluginkollektiv/assets/raw/master/cachify/images/cachify-toolbar-icon.png
https://www.pc-tweak.de/entwicklung/publish-remove.png
https://www.pc-tweak.de/entwicklung/settings.png

[#1 assets](https://github.com/pluginkollektiv/assets/pull/1)

Zodiac1978:
The idea is great, but the quality of the images is bad. Even the Icons in the dashboard are looking bad. 

Here are some screenshots from me:
![bildschirmfoto 2017-01-07 um 02 28 19](https://cloud.githubusercontent.com/assets/870475/21738155/0efd39b8-d481-11e6-90c6-b587ca172bbc.png)
![bildschirmfoto 2017-01-07 um 02 27 22](https://cloud.githubusercontent.com/assets/870475/21738157/0f07331e-d481-11e6-9a96-43cdf23b926b.png)
![bildschirmfoto 2017-01-07 um 02 27 10](https://cloud.githubusercontent.com/assets/870475/21738156/0f04f78e-d481-11e6-9002-52e0e23885fd.png)
![bildschirmfoto 2017-01-07 um 02 26 52](https://cloud.githubusercontent.com/assets/870475/21738158/0f094f14-d481-11e6-8135-4a09fc8cde2b.png)

timse201:
Yes i weboptimzed them to reduze size
Thought that s.o. could load it with mobile devices

websupporter:
Since #108 we have a new option in 2.3.0 on the settings page. Here an updated screenshot:
![bildschirmfoto am 2017-07-03 um 17 45 25](https://user-images.githubusercontent.com/6458412/27797771-7e601f42-6017-11e7-8f40-b7873aa69761.png)

krafit:
@websupporter yep, got this on my list. But I'm not entirely pleased with the current state of the backend. Some bits and pieces still need a little love, to get `2.3.0` out.

00travelgirl00:
How about this screenshot? The UI changed a bit already and it looks pretty smooth to me. 


[...]

PR #63, opened 2017-01-05
descriptions for new screenshots

timse201:
#57


Zodiac1978:
The last one could be optimized:
"Cache empty if articles have been edited"

maybe:
"Discard page/post cache or total cache if page/post is updated."

Issue #40, opened 2016-08-05
Convert icon font to SVG


swissspidy:
SVG files are considered superior nowadays compared to icon fonts.

See: https://github.com/pluginkollektiv/cachify/blob/7689f556ee0c85de554bbc2a865a2d59852d6422/css/dashboard.css#L5

See also #39.

2ndkauboy:
I totally agree. Does anyone knows how to extract the vector graphic from an opentype font? Using the SVG than instead of the font should be easy.

swissspidy:
Looks like there are some online font->svg converter tools around

PR #64, opened 2017-01-06
Convert icon font to SVG

websupporter:
Solves https://github.com/pluginkollektiv/cachify/issues/40

Solution:
Instead of using CSS pseudo classes, this PR adds a `<svg>` in the `<a>` element.

I also added a gruntfile, since I did a lot with the CSS during my patch and this was more easy for me, letting the minified file be created automatically. I think future development can benefit from this.


websupporter:
before-after

Here a quick before and after. The HDD upsite down was before 026d162 and is already fixed.

websupporter:
Well, well. I just see, the other two icons need to be upsite down as well. I will fix this, but let's see first, if my approach is one, we could follow.

Zodiac1978:
Looks like the paddings have to be adjusted a bit.

websupporter:
Thanks @Zodiac1978. Sorry for my design blindness: Where? :thinking:

timse201:
Memory should be more to the left
The hard drive 1px more to the left
Is the database smaller?

Issue #39, opened 2016-08-05
Properly use `wp_add_inline_style()`for the admin bar icon styling


swissspidy:
Currently the plugin prints out the CSS directly, which isn't that nice.

See: https://github.com/pluginkollektiv/cachify/blob/9132770ba5102bf2eba0ec41a0b453d5106fa9f2/inc/cachify.class.php#L708

See also #38.

visionlike:
I have testet with my changes and it works exactly in Frontend and Backend. With Valid HTML, which isn't that nice?

swissspidy:
Thanks again for opening the _PR_ as suggested by @2ndkauboy. I only closed it because it should have been an _issue_ instead.

Anyway, we'll look into it and perhaps even change it. At the same time I think it's not that big of a deal. Your visitors aren't going to see it, so their HTML will be totally valid. I also outlined my other concerns on #38.

visionlike:
I just wanted to help to keep it technically correct. I thought nothing of it. My Visitors are Web Developer, my Website is a sign. The Visitors can login to exchange and Work, that's all. But is ok, I do not care. I use it simply with my adaptation, the modification cost me no Minute.

I wish a pleasant evening.

Zodiac1978:
Same problem with these lines:
https://github.com/pluginkollektiv/cachify/blob/9132770ba5102bf2eba0ec41a0b453d5106fa9f2/inc/cachify.class.php#L1695-L1700

And the use of `valign` should be changed too:
https://github.com/pluginkollektiv/cachify/blob/9132770ba5102bf2eba0ec41a0b453d5106fa9f2/inc/cachify.class.php#L1713 (and more)

websupporter:
`valign` can be replaced without any problem. The vertical-align is already set from the WP stylesheet.

admin-bar: Since its a one liner, I think ( I guess we all agree ), it should be inline and not as a css file. I had a look. The thing is (dat ding is) WP prints out the adminbar styles as well (see `_admin_bar_bump_cb()` and how it is hooked into `wp_head`). So I can't find a proper handle to hook the `wp_add_inline_style()` in. So what would be the preferred way, if we do not have a handle?

With the inline on the options page. All it does is it sets the height of the input fields to 30px. The WP style is 28px. Haven't seen a dramatic change by just deleting the declarations and would suggest to do so.

Can prepare a PR, if someone could give me a hint with the admin_bar styles.

Thank you.

timse201:
[...]

Issue #32, opened 2016-07-12
General UI Optimizations


swissspidy:
Let's improve the accessibility and the overall UI of the plugin for a better UX.

Related: #30

timse201:
Think it would be nice to add:
- faq in the footer besides documentation or replace it because it is german only. the faq can be translated on translate.wordpress.org and the link can be changed there to the specific language page.
- add the current cache size also to the settings page
- inform the user that saving these settings will flush the cache
- add placeholders ( #53 )
- variable boxes ( #53 )

Issue #27, opened 2016-07-07
Provide .htaccess changes in the plugin


Zodiac1978:
If the user is using a caching method which needs a change of the .htaccess file, we should provide the code in the plugin itself for copy&paste.

Bonus points if we check if the server is providing correct data for %{DOCUMENT_ROOT} and provide a solution for this problem, see [Wiki](https://github.com/pluginkollektiv/cachify/wiki/Caching-Methoden#festplatte-hdd-cache)

swissspidy:
Even better: Directly write to the `.htaccess` file. I think WP Simple Cache does this quite well and simple.

websupporter:
Any thoughts here on NGINX. I do not have a lot of experience with it, but how does WordPress handle this in general and do we need something like a Nginx/Apache detection before providing a .htaccess/nginx configuration?

Zodiac1978:
If we can detect nginx we should provide (or write) the code for nginx directly.

websupporter:
Es gibt wohl eine globale `$is_nginx` definiert in `wp-includes/vars.php`

```
/**
 * Whether the server software is Apache or something else
 * @global bool $is_apache
 */
$is_apache = (strpos($_SERVER['SERVER_SOFTWARE'], 'Apache') !== false || strpos($_SERVER['SERVER_SOFTWARE'], 'LiteSpeed') !== false);

/**
 * Whether the server software is Nginx or something else
 * @global bool $is_nginx
 */
$is_nginx = (strpos($_SERVER['SERVER_SOFTWARE'], 'nginx') !== false);

/**
 * Whether the server software is IIS or something else
 * @global bool $is_IIS
 */
$is_IIS = !$is_apache && (strpos($_SERVER['SERVER_SOFTWARE'], 'Microsoft-IIS') !== false || strpos($_SERVER['SERVER_SOFTWARE'], 'ExpressionDevServer') !== false);

/**
 * Whether the server software is IIS 7.X or greater
 * @global bool $is_iis7
 */
$is_iis7 = $is_IIS && intval( substr( $_SERVER['SERVER_SOFTWARE'], strpos( $_SERVER['SERVER_SOFTWARE'], 'Microsoft-IIS/' ) + 14 ) ) >= 7;
```

Zodiac1978:
English, please! ;)

Zodiac1978:
English, please! ;) Nice finding. :)

timse201:
think we can use both
[...]

PR #69, opened 2017-01-26
implementation of config hints; some restructure; donation link

timse201:
what i changed:
* the ui is now more or less in separate files
* added a navbar
* paypal to donate; support link added; updated paypal link (as discussed on slack)
* deleted unnecessary valign and inline styles
* enhanced error message (the HTML code for the info symbol will be replaced by WordPress and a svg img is displayed)

to do list:
* nginx needs to be tested - will it include the right page? I have no test platform to do that
* your thoughts ;)

#27 #32 #39 #68


inc/setup/cachify.hdd.nginx.php
```diff
@@ -0,0 +1,68 @@
...
+    if ( $request_uri ~ /wp-admin/ ) {
+        return 405;
+    }
+    if ( $http_cookie ~ (wp-postpass|wordpress_logged_in|comment_author)_ ) {
+        return 405;
+    }
+
+    error_page 405 = @nocache;
+
+    try_files /wp-content/cache/cachify/${host}${uri}index.html @nocache;
```
timse201: path ok? or do we need to edit it to make use of content_url( $path ) ?

inc/setup/cachify.apc.nginx.php
```diff
@@ -0,0 +1,26 @@
...
+				</label>
+			</td>
+		</tr>
+	</table>
+
+	<div name="cachify[apc_nginx]" style="background:#fff;border:1px solid #ccc;padding:10px 20px"><pre>
+location ~ .php {
+    include fastcgi_params;
+    fastcgi_pass 127.0.0.1:9000;
+    fastcgi_param PHP_VALUE auto_prepend_file=/pfad/plugins/cachify/apc/proxy.php;
```
timse201: full path or relative path to plugins?

inc/setup/cachify.apc.htaccess.php
```diff
@@ -0,0 +1,18 @@
...
+				<label for="cachify_setup">
+					Please add the following lines to your .htaccess file
+				</label>
+			</td>
+		</tr>
+	</table>
+
+	<div name="cachify[apc_htaccess]" style="background:#fff;border:1px solid #ccc;padding:10px 20px"><pre>
+<Files index.php>
+    php_value auto_prepend_file /pfad/plugins/cachify/apc/proxy.php
```
timse201: full path or relative path to plugins?

inc/setup/cachify.hdd.nginx.php
```diff
@@ -0,0 +1,79 @@
...
+    if ( $request_uri ~ /wp-admin/ ) {
+        return 405;
+    }
+    if ( $http_cookie ~ (wp-postpass|wordpress_logged_in|comment_author)_ ) {
+        return 405;
+    }
+
+    error_page 405 = @nocache;
+
+    try_files /wp-content/cache/cachify/https-${host}${uri}index.html /wp-content/cache/cachify/${host}${uri}index.html @nocache;
```
timse201: content_url( $path ) needed?

inc/setup/cachify.hdd.nginx.php
```diff
@@ -0,0 +1,79 @@
...
+}
+
+## NOCACHE LOCATION
+location @nocache {
+    try_files $uri $uri/ /index.php?$args;
+}
+
+## PROTECT CACHE
+location ~ /wp-content/cache {
```
timse201: content_url( $path ) needed?

inc/setup/cachify.hdd.nginx.php
bueltge: have you tried this? Currently without testing I don't know.

timse201: You mean try files? I didnt test this. I have no nginx server. But this code was posted by swispydy on GitHub in an issue (sry I'm on a mobile device). And the user posted this code on his website so the code should work.
My thought behind all this is we should include standard configs that work for every server with less notes so a standard user is not confused. If someone wants specific code for HTTP or HTTPS only he can look into the wiki. But if we try to include 3 different configs it will confuse all users that are new with such things

bueltge: No, I mean have try with your idea to use `content_url( $path ) `.

timse201: Oh OK
No I need feedback if we need it here. What's the right path for these both code lines?
Added content URL in other files here, too. With my config it prints the relative path to WP content

bueltge:
Should great and helpful, but we need a test about different installs.

PR #73, opened 2017-03-26
Reformat code according to WordPress Coding Standards

herrbischoff:
Reference: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/

* Space usage adjustments.
* Whitespace removal.
* Formatting fixes.
* [Yoda Conditions](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#yoda-conditions).
* Strict comparisons.
* Change occurrences of `parse_url` to `wp_parse_url`.

There are still issue with the code that should be addressed like direct database calls, unescaped data, processing form data without nonce verification, `switch_to_blog` usage, super global usage, weird assignments and others — this is a start.

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1,85 +1,69 @@
 <?php
 /* Quit */
-defined('ABSPATH') OR exit;
-
+defined( 'ABSPATH' ) || exit;
 
 /**
 * Cachify
 */
-
 final class Cachify {
 
-
 	/**
-	* Plugin options
-	*
-	* @since  2.0
-	* @var    array
-	*/
-
+	 * Plugin options
+	 *
+	 * @since  2.0
+	 * @var    array
+	 */
 	private static $options;
 
-
 	/**
-	* Caching method
-	*
-	* @since  2.0
-	* @var    object
-	*/
-
+	 * Caching method
+	 *
+	 * @since  2.0
+	 * @var    object
+	 */
 	private static $method;
 
-
 	/**
-	* Method settings
-	*
-	* @since  2.0.9
-	* @var    integer
-	*/
-
+	 * Method settings
+	 *
+	 * @since  2.0.9
+	 * @var    integer
+	 */
 	const METHOD_DB = 0;
 	const METHOD_APC = 1;
 	const METHOD_HDD = 2;
 	const METHOD_MMC = 3;
 
-
 	/**
-	* Minify settings
-	*
-	* @since  2.0.9
-	* @var    integer
-	*/
-
+	 * Minify settings
+	 *
+	 * @since  2.0.9
+	 * @var    integer
+	 */
 	const MINIFY_DISABLED = 0;
 	const MINIFY_HTML_ONLY = 1;
 	const MINIFY_HTML_JS = 2;
 
-
 	/**
-	* Pseudo-Constructor
-	*
-	* @since   2.0.5
-	* @change  2.0.5
-	*/
-
-	public static function instance()
-	{
+	 * Pseudo-Constructor
+	 *
+	 * @since   2.0.5
+	 * @change  2.0.5
+	 */
+	public static function instance() {
 		new self();
 	}
 
-
 	/**
-	* Constructor
-	*
-	* @since   1.0.0
-	* @change  2.2.2
-	*
-	* @param   void
-	* @return  void
-	*/
-
-	public function __construct()
-	{
+	 * Constructor
+	 *
+	 * @since   1.0.0
+	 * @change  2.2.2
+	 *
+	 * @param   void
+	 * @return  void
+	 */
+	public function __construct() {
 		/* Set defaults */
 		self::_set_default_vars();
 
@@ -88,7 +72,7 @@ final class Cachify {
 			'init',
 			array(
 				__CLASS__,
-				'register_publish_hooks'
+				'register_publish_hooks',
 			),
 			99
 		);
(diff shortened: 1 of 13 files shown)


websupporter:
Hi, we wanted to use also `wp_parse_url()` and got this friendly reminder:
> According to documentation, wp_parse_url is available from WordPress 4.4, so required WP version (3.8 in the moment) should be bumped before this is merged.

https://github.com/pluginkollektiv/antispam-bee/pull/85

herrbischoff:
Feel free to review the code and annotate whatever should be changed or added.

websupporter:
I will have a closer look @herrbischoff,
but first of all thanks for this great patch! :tada:

herrbischoff:
@websupporter: the `wp_parse_url` are [in here](https://github.com/pluginkollektiv/cachify/pull/73/commits/351c48a49daecfacb81dca0cccba3a3d3313bfc6#diff-757474562486eb6b807e4a5f3257905eL504).

inc/cachify.settings.php
```diff
@@ -61,14 +61,14 @@
 			<td>
 				<fieldset>
 					<label for="cachify_without_ids">
-						<textarea name="cachify[without_ids]" id="cachify_without_ids" rows="1" placeholder="<?php esc_attr_e( 'e.g. 1,2,3', 'cachify' ); ?>"><?php echo esc_attr($options['without_ids']) ?></textarea>
+						<textarea name="cachify[without_ids]" id="cachify_without_ids" rows="1" placeholder="<?php esc_attr_e( 'e.g. 1,2,3', 'cachify' ); ?>"><?php echo esc_attr( $options['without_ids'] ) ?></textarea>
```
websupporter: Is there a reason, why we do not use `esc_textarea()`?

inc/cachify.settings.php
```diff
@@ -61,14 +61,14 @@
...
 						<?php esc_html_e( 'Post/Pages-IDs', 'cachify' ); ?>
 					</label>
 
 					<br />
 
 					<label for="cachify_without_agents">
-						<textarea name="cachify[without_agents]" id="cachify_without_agents" rows="1" placeholder="<?php esc_attr_e( 'e.g. MSIE 6, Opera', 'cachify' ); ?>"><?php echo esc_attr($options['without_agents']) ?></textarea>
+						<textarea name="cachify[without_agents]" id="cachify_without_agents" rows="1" placeholder="<?php esc_attr_e( 'e.g. MSIE 6, Opera', 'cachify' ); ?>"><?php echo esc_attr( $options['without_agents'] ) ?></textarea>
```
websupporter: see above.

inc/cachify.class.php
```diff
@@ -1694,75 +1563,68 @@ public static function validate_options($data)
...
+					}
+					echo '</h2>';
+
+					switch ( $current_tab ) {
+						case 'settings' :
+							include 'cachify.settings.php';
+							break;
+						case 'setup' :
+							if ( self::METHOD_HDD === $options ['use_apc'] ) {
+								if ( $is_nginx ) { include 'setup/cachify.hdd.nginx.php';
```
websupporter: I just realize, we do not import `$is_nginx` and it is undefined. Would be great, if we could import it in this patch.

inc/setup/cachify.apc.htaccess.php
```diff
@@ -24,10 +24,10 @@
 	</table>
 
 	<div style="background:#fff;border:1px solid #ccc;padding:10px 20px">
-		<pre><?php echo sprintf ( '%s%s%s',
-				$beginning,
-				WP_PLUGIN_DIR,
-				$ending ); ?></pre>
+		<pre><?php echo sprintf( '%s%s%s',
+			$beginning,
```
websupporter: Do we want to use late escaping?

inc/setup/cachify.apc.htaccess.php
```diff
@@ -24,10 +24,10 @@
...
+			WP_PLUGIN_DIR,
+		$ending ); ?></pre>
```
websupporter: see above.

inc/cachify.class.php
```diff
@@ -1694,75 +1563,68 @@ public static function validate_options($data)
...
+									'setup' => esc_html__( 'Setup', 'cachify' ),
+							);
+
+					$current_tab = isset( $_GET['cachify_tab'] ) ? $_GET['cachify_tab'] : 'settings';
+
+					echo '<h2 class="nav-tab-wrapper">';
+					foreach ( $cachify_tabs as $tab => $name ) {
+						$class = ($tab === $current_tab) ? ' nav-tab-active' : '';
+						$link = "?page=cachify&cachify_tab=$tab";
+						echo "<a class='nav-tab$class' href='$link'>$name</a>";
```
websupporter: Do we want to use late escaping?

websupporter:
For me, this patch looks very good. Thanks a lot. It raises the question, whether we want to drop version support for 3.8 - 4.3. I try to get some feedback on this issue.

The advantage of using `wp_parse_url` is, this
> function that handles consistency in the return values across PHP versions

So, it would make us more independent from the PHP version we are running on :tada:

I had another question, raised by this PR but not introduced by it:
Should we use late escaping like a standard? Right now, everywhere where it is not done, there is no harm, but on the long run, I can change variables without the need to check, whether they are escaped in the end or not, because I can rely on the fact, the variable will be escaped, when it comes to output it. But to me, we could handle this also in another Issue/PR.

`$is_nginx`
I also saw, while doing me review, The `$is_nginx` is undefined. We should define it. Here the link to the current master: https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.class.php#L1750
If we do not handle this issue in this PR, we should open up an issue to keep track.

My last comment is about some usage of `esc_attr()` where `esc_textarea()` might be more appropriate (https://github.com/pluginkollektiv/cachify/pull/73/files#r108593354). Its a bit picky and not a real stopper, but I thought I mention it.

So, the crucial question is: Do we drop support?

herrbischoff:
I'm all for dropping support for older WordPress versions. There is really no reason to support officially abandoned versions with known security vulnerabilities.

krafit:
Yep, let's do it 👍

Issue #76, opened 2017-03-30
$is_nginx not defined


websupporter:
https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.class.php#L1750

We need to do a `global $is_nginx` here.

PR #77, opened 2017-03-30
Add global $is_nginx (Solves #76)

websupporter:
Solves #76 by declaring the global $is_nginx.


bueltge:
@websupporter Maybe we should use a var inside the class, like `$this->is_nginx = $GLOBALS['is_nginx'];` so that we only use the global inside the constructor?

websupporter:
☝️ :smile:

herrbischoff:
Looks good.

PR #83, opened 2017-04-05
Fix undefined variable in Cachify::options_page().

### LINKED ISSUE #82: Variable $is_nginx is undefined
It seems that variable `$is_nginx` [here](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.class.php#L1633) should in fact be `self::$is_nginx`.

chesio:
Fixes #82.


websupporter:
Thanks a lot @chesio for catching this! :tada:

chesio:
@websupporter Actually, my IDE caught this... :-)

PR #91, opened 2017-05-08
Translate German residues into English

chesio:
Only comments and (one) private method name are affected.


krafit:
thanks, @chesio

PR #93, opened 2017-05-11
Print cached data earlier. Fixes #92.

### LINKED ISSUE #92: Output cached data earlier
Hello pluginkollektiv team,

I have a small optimization suggestion for DB cache. Currently, Cachify checks whether there is anything in DB cache in [Cachify::manage_cache()](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.class.php#L1292-L1321) method that is being hooked in `template_redirect` action, ie. very late.

I propose to move this check right into Cachify constructor that is being called from `plugins_loaded` action. Of course, this would mean that [Cachify::_skip_cache()](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.class.php#L1107-L1156) test would no longer affect cache retrieval (only cache creation), but I don't see a problem here, because this is exactly how all the other caching methods work.

The change is simple to implement, I'll submit PR to illustrate what I mean. I run some tests on a very simple website and I got 50% performance gain (both in number of DB queries and execution time).

chesio:
Cache data retrieval and output is done in separate method that is called early (see #92).

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -75,6 +75,11 @@ final class Cachify {
 		/* Set defaults */
 		self::_set_default_vars();
 
+		if ( ! is_admin() ) {
+			/* Check whether there are cached data to be printed. */
+			self::print_cache();
+		}
+
 		self::$is_nginx = $GLOBALS['is_nginx'];
 
 		/* Publish hooks */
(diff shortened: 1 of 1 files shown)


chesio:
Ok, I just realized that I was too hasty with this PR, because some of the checks in Cachify::_skip_cache() method have to be applied for cache data retrieval too (like `empty( $_GET )` etc.) I'll add them tomorrow.

chesio:
Tomorrow or actually later today :-)

chesio:
Hi @bueltge, thanks for merging this PR so promptly, but in fact I wanted to add one more commit to it :-) But no problem, I submitted it via new PR instead (see #95).

bueltge:
Really welcome, thanks again for your work on the plugin.

PR #98, opened 2017-05-20
PHP fallback for HDD Cache

### LINKED ISSUE #97: PHP fallback for HDD caching method
Currently, if .htaccess rules for HDD caching are missing, HDD caching is effectively inactive, because `Cachify_HDD::print_cache()` method [does nothing](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify_hdd.class.php#L108-L110). I think that now, when [cached data are output early](https://github.com/pluginkollektiv/cachify/issues/92), HDD cache could have PHP fallback.

chesio:
Kicks in when .htaccess rules for HDD cache are missing. Fixes #97.

### DIFF
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -103,10 +103,14 @@ final class Cachify_HDD {
 	 * Print the cache
 	 *
 	 * @since   2.0
-	 * @change  2.0
+	 * @change  2.3
 	 */
 	public static function print_cache() {
-		return;
+		$size = @readfile( self::_file_html() );
+		if ( ! empty ( $size ) ) {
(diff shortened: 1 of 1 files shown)


chesio:
I saw `@` operator being (mis)used in several other places in plugin code, so I didn't bother, but certainly it should be avoided here.

This said, I think some general clean up of `@` operator occurences in plugin could be done - I'll create an issue for that.

bueltge:
Yes, you have right. The `@` operator is often fast helpful, usage in the plugin. However in my opinion not the best style for feature code, so that we should reduce this in our code, also in this plugin. A clean up would be really fine, but currently I haven't a idea of the effort.

Issue #108, opened 2017-06-27
Unify and cleanup HTML signature


stklcode:
Cachify provides a signature as HTML comment below the cached markup. This signature is uniform for HDD, Memcached and APC, but the DB cache shows a different format with additional information.

<details> 
<summary>Example HDD cache</summary>
<pre>
&lt;!-- Cachify | http://cachify.de
HDD Cache @ 01.01.2017 01:23:45 -->
</pre>
</details>

<details> 
<summary>Example DB cache</summary>
<pre>
&lt;!--
Cachify | http://cachify.de
Without Cachify: 50 DB queries, 0.10 seconds, 23,45 MB
With Cachify: 20 DB queries, 0.05 seconds, 12,34 MB
Generated: 10 hours ago
-->
</pre>
</details>

<br/>

**I got two points here:**

1. I don't see any real advantage of calculating the timespan for the DB backend, despite "because we can", so my suggestion is to unify the strings.
2. I assume that >99% of the page visitors don't care about caching at all and also most administrators don't check the query/memory details regularily. Details are OK for setup and debugging, but most time pointless in production environments. (some might also claim an information leak)
So my second suggestion is to make those details optional and print only minimal information.

With both changes, signatures would look like following

**Signature with details (old behavior, default?)**
```
<!-- Cachify | http://cachify.de
{METHOD} @ {DATE TIME}
{DETAILS}
-->
```

**Cleaned signature:**
```
<!-- Cachify | http://cachify.de
Generated @ {DATE TIME} -->
```

Open for discussion here. Would those changes be welcome?

If so, I got a branch ready for PR ( stklcode/cachify@31ea8ed6e6ec653ce89b6dd03facbd6f54860b30, stklcode/cachify@42c7ab99f36c5901dc376c9bf391633559b4c746 ).

krafit:
Thank you, @stklcode! You are right, I too see no point in providing additional information for DB caches. 
Please go ahead and open a PR.

stklcode:
Just one point I'm curious about. Right now the default behavior is the old (full) signature.

I didn't want to modify this without feedback, but swapping the default is obviously only one character at two positions ;)

krafit:
imho the new (slim) signature should be default

stklcode:
I totally agree. Swapped the default (and restored the changes on the settings page, I accidentially reverted...)

swissspidy:
Why even add anything to the HTML footer at all? Doesn't an HTTP header suffice?

[...]

PR #109, opened 2017-06-27
Unify and cleanup HTML signature

stklcode:
Addresses issue #108 

Unifies signature in HTML comment and provides the ability to remove additional data.

**Signature with details (old behavior)**
```
<!-- Cachify | http://cachify.de
{METHOD} @ {DATE TIME}
{DETAILS}
-->
```

**Cleaned signature (default)**
```
<!-- Cachify | http://cachify.de
Generated @ {DATE TIME} -->
```

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1026,7 +1026,7 @@ final class Cachify {
 	 * @return  bool  Show details in signature.
 	 */
 	private static function _signature_details() {
-		return isset(self::$options['sig_detail']) ? self::$options['sig_detail'] == 1 : false;
+		return self::$options['sig_detail'] === 1;
 	}
 
 	/**
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -475,7 +475,7 @@ final class Cachify {
 	 * Get options
 	 *
 	 * @since   2.0
-	 * @change  2.1.2
+	 * @change  2.3.0
 	 *
 	 * @return  array  Array of option values
 	 */
@@ -490,6 +490,7 @@ final class Cachify {
 				'without_agents' 	=> '',
 				'use_apc'		 	=> self::METHOD_DB,
 				'reset_on_comment'  => 0,
+				'sig_detail'        => 0,
 			)
 		);
 	}
@@ -1025,7 +1026,7 @@ final class Cachify {
 	 * @return  bool  Show details in signature.
 	 */
 	private static function _signature_details() {
-		return isset(self::$options['sig_detail']) ? self::$options['sig_detail'] == 1 : true;
+		return isset(self::$options['sig_detail']) ? self::$options['sig_detail'] == 1 : false;
 	}
 
 	/**
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1018,6 +1018,17 @@ final class Cachify {
 	}
 
 	/**
+	 * Determine if cache details should be printed in signature
+	 *
+	 * @since   2.3.0
+	 *
+	 * @return  bool  Show details in signature.
+	 */
+	private static function _signature_details() {
+		return isset(self::$options['sig_detail']) ? self::$options['sig_detail'] == 1 : true;
+	}
+
+	/**
 	 * Get hash value for caching
 	 *
 	 * @since   0.1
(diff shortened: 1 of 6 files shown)


inc/cachify.class.php
```diff
@@ -1018,6 +1019,17 @@ private static function _cache_expires() {
...
 
 	/**
+	 * Determine if cache details should be printed in signature
+	 *
+	 * @since   2.3.0
+	 *
+	 * @return  bool  Show details in signature.
+	 */
+	private static function _signature_details() {
+		return isset(self::$options['sig_detail']) ? self::$options['sig_detail'] == 1 : false;
```
websupporter: I think, this could be a bit more readable. Isn't `isset(self::$options['sig_detail'])` always `true`?

inc/cachify.class.php
```diff
@@ -1290,7 +1302,8 @@ public static function set_cache( $data ) {
 			),
 			self::_cache_hash(),
 			self::_minify_cache( $data ),
-			self::_cache_expires()
+			self::_cache_expires(),
+			self::_signature_details()
```
websupporter: Why is it needed here? `Cachify_DB::store_item` does not expect this information.

inc/cachify_db.class.php
```diff
@@ -131,7 +132,7 @@ public static function print_cache( $cache ) {
 
 		/* Signature */
 		if ( isset( $cache['meta'] ) ) {
-			echo self::_cache_signature( $cache['meta'] );
+			echo self::_cache_signature( $sigDetail, $cache['meta'] );
```
websupporter: What is the reason to store the signature in APC/HDD/Memcache in the cached page itself while DB is generating the signature during the output process? I think, this would make sense to unify.

inc/cachify_apc.class.php
```diff
@@ -33,13 +33,14 @@ public static function stringify_method() {
...
-	 * @change  2.0
+	 * @change  2.3.0
 	 *
-	 * @param   string  $hash      Hash of the entry.
-	 * @param   string  $data      Content of the entry.
-	 * @param   integer $lifetime  Lifetime of the entry.
+	 * @param   string  $hash       Hash of the entry.
+	 * @param   string  $data       Content of the entry.
+	 * @param   integer $lifetime   Lifetime of the entry.
+	 * @param   bool    $sigDetail  Show details in signature.
```
websupporter: Please use snake_case format: `$sig_detail`

inc/cachify.class.php
websupporter: Also about Yoda you must think and strict comparison you must use :)

inc/cachify_db.class.php
```diff
@@ -160,37 +161,51 @@ public static function get_stats() {
 	 * Generate signature
 	 *
 	 * @since   2.0
-	 * @change  2.0.5
+	 * @change  2.3.0
 	 *
-	 * @param   array $meta  Content of metadata.
-	 * @return  string       Signature string
+	 * @param   bool  $detail  Show details in signature
```
websupporter: Parameter comment must end with a full stop.

inc/cachify_db.class.php
```diff
@@ -160,37 +161,51 @@ public static function get_stats() {
...
-				self::_page_queries(),
-				self::_page_timer(),
-				self::_page_memory()
-			),
-			sprintf(
-				'Generated: %s ago',
-				human_time_diff( $meta['time'], current_time( 'timestamp' ) )
-			)
-		);
+		if ($detail) {
```
websupporter: space after opening bracket and before closing bracket are missing.

websupporter:
Hi @stklcode,
thanks a lot for your PR :tada: I have just some minor nitpicks about it (see above). Most of them just so we stay on track with our goal to align the code with WordPress standards.

inc/cachify_db.class.php
stklcode: The DB cache signature can provide additional information only available at runtime (e.g. number of actual DB queries). There might be any reason why one would like to see this...

Only possibility to unify call of the "basic" signature would be to add a second comment at runtime or drop the additional data completely (remove `$sig_detail`).

websupporter: thanks for clarification :)

PR #117, opened 2017-07-01
Reimplement settings screen output.

### LINKED ISSUE #110: Remove redundant settings footer
There are multiple instances of the same `<table>` element in the footer of each an every `/inc/setup/` file.

This is a maintenance nightmare 🙈

chesio:
Fixes #110 and by the way:

* Part of `Cachify::options_page()` method has been refactored into separate method: I think it makes sense to keep the logic that determines what tabs are applicable in context of current options separated from the method that renders the options page.
* The "setup" tabs are no longer wrapped in `<form></form>` tags.

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1611,7 +1611,7 @@ final class Cachify {
 	 */
 	public static function options_page() {
 		$options = self::_get_options();
-		$cachify_tabs = self::_get_tabs($options);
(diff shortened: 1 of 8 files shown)


krafit:
Thank you @chesio <3

This PR brings us one step closer to our upcoming 2.3 beta1 🎉

chesio:
Glad I could help :-) Looking forward to version 2.3!

PR #119, opened 2017-07-03
Update readme

krafit:
- WP 4.4 is now minimum requirement
- tested up to 4.8
- new PayPal donate links


chesio:
I believe the donate link in [settings page footer](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.settings_footer.php) needs to be updated too.

krafit:
@chesio yes, I already did this and some other footer tweaks locally, will commit them later

chesio:
@krafit Sorry, I noticed only now that this change relates to README file only.

PR #126, opened 2017-07-04
Fix: Cachify::_is_index() should return true, if script name is index.php

chesio:
Makes method name and phpdoc consistent with what the method does.


chesio:
Closing in favor of #127.

bueltge:
@chesio please you should add a entry in the chnagelog, so that we can use it for the next release. Thx.

chesio:
@bueltge Will do!

PR #128, opened 2017-07-04
fix #118 unify parse_url() usage

### LINKED ISSUE #118: Unfiy parse_url() usage
In cachify.class.php we are using `wp_parse_url()` while in cachify_hdd.class.php and cachify_memcache.class.php we are using `parse_url()`. We should unify the usage to `wp_parse_url()`

stklcode:
Quick fix for #118.

* remove the `$component` parameter from `wp_parse_url()` calls, as it's not supported < 4.7
* replace every `parse_url()` call with `wp_parse_url()` (4.4 compatible)

### DIFF
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -303,19 +303,16 @@ final class Cachify_HDD {
 	private static function _file_path( $path = null ) {
 		$prefix = is_ssl() ? 'https-' : '';
 
+		$host_parts = wp_parse_url( 'http://' . strtolower( $_SERVER['HTTP_HOST'] ) );
+		$path_parts = wp_parse_url( $path ? $path : $_SERVER['REQUEST_URI'] );
+
 		$path = sprintf(
 			'%s%s%s%s%s',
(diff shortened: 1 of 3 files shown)


inc/cachify_hdd.class.php
```diff
@@ -303,19 +303,16 @@ public static function _dir_size( $dir = '.' ) {
 	private static function _file_path( $path = null ) {
 		$prefix = is_ssl() ? 'https-' : '';
 
+		$host_parts = wp_parse_url( 'http://' . strtolower( $_SERVER['HTTP_HOST'] ) );
```
chesio: I think this can be skipped and you can use `$_SERVER['HTTP_HOST']` directly (the same way as in `Cachify_MEMCACHED::_file_path()` method.

stklcode: Absolutely right. My commit didn't optimize any existing code ;)

If we're talking about tweaks, `cachify.class.php` lines 1045 ff. could also be simplified, as `wp_parse_url( '' ) ['path'] === ''`, so the `empty()` check is redundant...

PR #132, opened 2017-07-06
Change admin notice to warning

### LINKED ISSUE #122: Check admin notification
I might be wrong on this one, but there is a kind of guideline for these admin nags (or our friend Caspar wrote about it) and we should check them out and see if this nag complies.

bildschirmfoto 2017-07-03 um 17 50 16

krafit:
Added .notice-warning to admin notice and removed glyph.
Fixes #122

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1597,11 +1597,8 @@ final class Cachify {
 			add_settings_error(
 				'cachify_method_tip',
 				'cachify_method_tip',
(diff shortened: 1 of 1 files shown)


krafit:
@websupporter yep, I stumbled across the same docs issue.
But since `update` and `error` are deprecated and the new classes work es well, I decided to ignore that glitch in the ~matrix~ documentation 😄 
Thanks for your review

Issue #135, opened 2017-07-12
Minimizing doesn’t work on my nginx with APC


MatzeKitt:
I use a fresh Debian Stretch system with the official builds of nginx and PHP 7. My caching method is APC and it does cache some files (if I may trust the stats on the dashboard so far).

The problem is that neither the HTML nor the JavaScript minimizing is working on my system. A live demo with the current active setting “HTML + Inline JavaScript” results in a source code as usual, uncompressed:
https://kittpress.com

I left every other setting untouched.

The nginx configuration is, of course, identical to the one in the wiki.

If you need any further information, please let me know.

stklcode:
Is your live demo up to date with Cachify still active? Looks like caching does not work at all.

If it worked, the HTML code should contain the Cachify signature in the end. Should look lilke this:
```
<!-- 
Cachify | http://cachify.de
APC Cache | 01.01.1970 12:34:56
-->
```

Did you add the corresponding lines to your @php.ini@ (or PHP config within nginx), as described on the "Setup" tab of the Cachify settings?

KittBlog:
Yes, it is.

And yes, I added the PHP configuration inside of nginx:
```
        ##
        # Cachify
        ##

        location ~ .php {
                include /etc/nginx/fastcgi.conf;
                fastcgi_pass 127.0.0.1:9001;
                fastcgi_param PHP_VALUE auto_prepend_file=/var/www/kittpress/wp-content/plugins/cachify/apc/proxy.php;

                location ~ /wp-admin/ {
                        include /etc/nginx/fastcgi.conf;
                        fastcgi_pass 127.0.0.1:9001;
                        fastcgi_param PHP_VALUE auto_prepend_file=;
                }
        }
```

You can ignore the different filename for `fastcgi_params` as I use this file for a custom implementation for another software.

However, I can’t find any setting I should change in the `php.ini` file. Did I miss something here?

I noticed this:
https://wordpress.org/support/topic/plugin-has-no-valid-headerapc-caching/#post-8698598

So if this is still the current state of the plugin it can’t be used within PHP 7 because there is no more APC but OPcache and APCu.

stklcode:
> However, I can’t find any setting I should change in the `php.ini` file. Did I miss something here?

No, the description is just little confusing. You might as well place the `auto_prepend_file` value inside the `php.ini` (global or pool), if not passed by nginx. Should give the same result

[...]

PR #137, opened 2017-07-12
Fix hash function for APC proxy

stklcode:
The added prefix for HTTPS pages, introduced in #2, has not been added for the APC proxy which breaks APC functionallity for HTTPS pages (reported in #135).

Just copied the `is_ssl()` from WP to be consistent and adapted the hash function. Didn't test though...

### DIFF
--- a/apc/proxy.php
+++ b/apc/proxy.php
@@ -9,12 +9,35 @@ if ( ! empty( $_COOKIE ) ) {
 	}
 }
 
+/**
+ * Determines if SSL is used.
+ *
+ * @see is_ssl() (wp-includes/load.php).
+ *
+ * @return bool True if SSL, otherwise false.
+ */
+function cachify_is_ssl() {
+	if ( isset( $_SERVER['HTTPS'] ) ) {
+		if ( 'on' == strtolower( $_SERVER['HTTPS'] ) ) {
+			return true;
+		}
+
+		if ( '1' == $_SERVER['HTTPS'] ) {
+			return true;
+		}
+	} elseif ( isset( $_SERVER['SERVER_PORT'] ) && ( '443' == $_SERVER['SERVER_PORT'] ) ) {
+		return true;
+	}
+
+	return false;
+}
+
 if (
 	empty( $_cachify_logged_in )
 	&& ( strpos( filter_input( INPUT_SERVER, 'PHP_SELF', FILTER_SANITIZE_STRING ), '/wp-admin/' ) === false )
 	&& ( strpos( filter_input( INPUT_SERVER, 'HTTP_ACCEPT_ENCODING', FILTER_SANITIZE_STRING ), 'gzip' ) !== false )
 	&& extension_loaded( 'apc' )
-	&& ( $cache = apc_fetch( md5( $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'] ) . '.cachify' ) )
+	&& ( $cache = apc_fetch( md5( ( cachify_is_ssl() ? 'https-' : '' ) . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'] ) . '.cachify' ) )
 ) {
 	ini_set( 'zlib.output_compression', 'Off' );
 
--- a/apc/proxy.php
+++ b/apc/proxy.php
@@ -18,14 +18,14 @@ if ( ! empty( $_COOKIE ) ) {
  */
 function cachify_is_ssl() {
 	if ( isset( $_SERVER['HTTPS'] ) ) {
-		if ( 'on' == strtolower( $_SERVER['HTTPS'] ) ) {
+		if ( 'on' === strtolower( $_SERVER['HTTPS'] ) ) {
 			return true;
 		}
 
-		if ( '1' == $_SERVER['HTTPS'] ) {
+		if ( '1' === $_SERVER['HTTPS'] ) {
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


apc/proxy.php
```diff
@@ -9,12 +9,35 @@
...
+/**
+ * Determines if SSL is used.
+ *
+ * @see is_ssl() (wp-includes/load.php).
+ *
+ * @return bool True if SSL, otherwise false.
+ */
+function cachify_is_ssl() {
+	if ( isset( $_SERVER['HTTPS'] ) ) {
+		if ( 'on' == strtolower( $_SERVER['HTTPS'] ) ) {
```
bueltge: We should check strict, like `===` to create a solid code base. Would you change this, enhance this to more scrict source?

stklcode: I'm not sure if that's what we want here. I'm 100% with you in general, but ...

1. The method is copy&paste from the original WP method (used to generate the hash). I intentionally did not change the code for consistency with both calls
2. I really does not make any difference here, as `strtolower()` always returns a string.

If you still want this, I'll change it, but I don't see any benefit here and the reason is stated in the PHPDoc...

bueltge: I had seen, that its a copy from the WP source, but maybe this function inside the core is also not the best. My point for this strict coding is performance. The `==` coerces or converts the data type temporarily. The strict comparison is much faster and if we can get sure, that the string is right, then we should use the strict comparison.
My two cents, maybe the other will get more values on your reason, there is also valid.

stklcode: **tl;dr**  We need a third opinion to decide.

> The strict comparison is much faster 

Yes it is what's propagated all over the web and true in many cases, but not always. Run a simple (primitive) benchmark with strings yourself:
<details> 
<summary>the script</summary>
<pre>&lt;?php
$test = 'foobar';
$max = 100000000;

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
  if('foobar' == $test) {};
}
$end = microtime(true);
print 'matching     ==  ' . ($end-$start) . "\n";

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
  if('foobar' === $test) {}
}
$end = microtime(true);
print 'matching     === ' . ($end-$start) . "\n";

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
          if('barfoo' != $test) {};
}
$end = microtime(true);
print 'matching     !=  ' . ($end-$start) . "\n";

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
          if('barfoo' !== $test) {}
}
$end = microtime(true);
print 'matching     !== ' . ($end-$start) . "\n";

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
  if('barfoo' == $test) {}
}
$end = microtime(true);
print 'non-matching ==  ' . ($end-$start) . "\n";

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
  if('barfoo' === $test) {}
}
$end = microtime(true);
print 'non-matching === ' . ($end-$start) . "\n";

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
  if('foobar' != $test) {}
}
$end = microtime(true);
print 'non-matching !=  ' . ($end-$start) . "\n";

$start = microtime(true);
for($i = 0; $i < $max; $i++) {
  if('foobar' !== $test) {}
}
$end = microtime(true);
print 'non-matching !== ' . ($end-$start) . "\n";</pre>
</details>

My results (PHP 7.0.21 CLI, 100mio runs) show a tony difference in the opposite direction
```
matching     ==  1.2242779731751
matching     === 1.3983459472656
matching     !=  1.5629131793976
matching     !== 1.7631139755249
non-matching ==  1.7444911003113
non-matching === 1.8354380130768
non-matching !=  1.1368229389191
non-matching !== 1.5221791267395
```
>  if we can get sure, that the string is right, then we should use the strict comparison

If we don't trust the output of a PHP core method, strict comparison is our least problem. The input is not user provided here.

Please don't get me wrong, I don't want to argue about peanuts here, but from my experience, a clean "clone management" has much more benefit for maintaining integrity than adapting everything to own conventions.

krafit: I'd like to hear @2ndkauboy's opinion on this!

bueltge:
@stklcode no worries, I like your effort on this discussion. 
I think always a type accesa and strict coding standard is much easier to maintain and get helpful error messages. Yes, the WP core haven't this, but the core have a lot of old ideas. Your values for speed are interesting. I mean I have also repository with results in different php versions. 
However really thank you for your examples, your effort and your heart to work this.

stklcode:
No problem @bueltge. Always interesting what kind of discussions derive from alleged trivialities.

The performance actually changed dramatically from PHP 5 to 7. In 5.6, `===` is ~5% faster (overall 200% slower, but that's a different story). Internal management of the type conversion has been improved here.

Let's just wait for the next statement and merge, whatever results. Changing to `===` is a 10s fix after all...

bueltge:
Also to. The fix is created fast. But maybe we get a final conclusion, usable for all repos and next changes, improvements in our plugin context.

apc/proxy.php
2ndkauboy: I don't see a huge performance deficit or benefit here (the strtolower function also adds some runtime). But honestly, this is probably a time in nanoseconds and doesn't really sums up especially when called only once per request.

I would go for the strict comparison only because it's recommended/necessary due to the strict WordPress Coding Standards. And as the header will always be a string, the strict comparison should not break anything.

krafit: Thank you @2ndkauboy ❤️

stklcode: Thanks. 2/3 => changed it.

PR #142, opened 2017-07-21
Add composer.json to the project.

### LINKED ISSUE #111: Add composer.json file to the plugin
Hi,

Having `composer.json` file in GitHub repo would make it easier to fetch updates from master branch via Composer. I'm currently running master branch in several of my projects, because Cachify version in WordPress Plugin Directory (and thus [WordPress Packagist](https://wpackagist.org/) as well) is a bit outdated in the moment...

chesio:
Fixes #111. I think there's no harm in having `composer.json` in git repo.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,28 @@
+{
+  "name": "pluginkollektiv/cachify",
+  "type": "wordpress-plugin",
+  "license": "GPL 2.0",
+  "description": "Smart but efficient cache solution for WordPress. Use DB, HDD, APC or Memcached for storing your blog pages. Make WordPress faster!",
+  "homepage": "https://wordpress.org/plugins/cachify/",
+  "authors": [
+    {
+      "name": "Sergej Müller",
+      "homepage": "https://sergejmueller.github.io"
+    },
+    {
+      "name": "pluginkollektiv",
+      "homepage": "http://pluginkollektiv.org"
+    }
+  ],
+  "keywords": [
+    "wordpress", "wordpress-plugin"
+  ],
+  "support": {
+    "issues": "https://github.com/pluginkollektiv/cachify",
+    "forum": "https://wordpress.org/support/plugin/cachify"
+  },
+  "require": {
+    "php": ">=5.2.0",
+    "composer/installers": "~1.0"
+  }
+}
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "pluginkollektiv/cachify",
+  "type": "wordpress-plugin",
+  "license": "GPL 2.0",
```
stklcode: Actually Cachify is licensed unter GPLv2 **or later**, so according to SPDX, the correct license string should be `GPL-2.0+`. Even if we'd omit the +, it should still be `GPL-2.0`.

stklcode:
+1 for that. Looks good overall.

However we should also add `vendor` and the `composer.lock` to `.gitignore` (or commit the .lock, but I prefer excluding it)

composer.json
chesio: Thanks for feedback, I wasn't aware of it. It's fixed!

chesio:
> (or commit the composer.lock, but I prefer excluding it)

May I ask why? I'm asking out of curiosity, I'm not seasoned Composer user, so I only know that official documentation [recommends commiting](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control) `composer.lock` ([except for libraries](https://getcomposer.org/doc/02-libraries.md#lock-file)). Is there any good reason not to?

stklcode:
That's right for projects depending on third-party packages/libs. We do not have any real dependencies here required for distribution at the moment. And even if we add PHPUnit or PHPCS to the dev-dependencies the precise versions do not really matter (OK, maybe for code-style definitions). From that point, the `.lock` does not hurt, but is not important for now either.

Second point is working with PHPUnit over composer (we don't do yet) with different PHP versions supported from 5.2. Can be really annoying when developing with different local versions, if someone locked on 6.x.x (compatible >= 7.0) and you develop with 5.x or inverse, locking to PHPUnit 4 which breaks with 7... Of course you can locally override or use an external lib, but it's extra work and we're lazy :smirk:

TL;DR don't see a benefit from commiting this right now, but add it, if you like.

chesio:
> TL;DR don't see a benefit from commiting this right now, but add it, if you like.

Me neither, I was mainly interested in what you have to say on this. Thanks for explanation!

chesio:
I've added `composer.lock` and `vendor` to `.gitignore`.

PR #144, opened 2017-08-24
Allow HDD cache to be cleared when cachify_user_can_flush_cache filter returns true for current user.

chesio:
I want all users to be able to clear Cachify cache, so I'm using the following hook:
```php
add_filter('cachify_user_can_flush_cache', '__return_true');
```

Clear cache icon is shown, admin notice says cache is cleared, but no files get deleted from disk. The problem is that there's an extra check done before any file gets physically deleted and it has not been filtered.

### DIFF
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -363,7 +363,7 @@ final class Cachify_HDD {
 			return false;
 		}
 
-		if ( ! current_user_can( 'manage_options' ) ) {
+		if ( ! apply_filters( 'cachify_user_can_flush_cache', current_user_can( 'manage_options' ) ) ) {
 			return false;
 		}
 


bueltge:
@chesio Please enhance the readme about this new filter hook, so that user get information about the hook. Thx.

chesio:
@bueltge I found out that hooking into the `cachify_user_can_flush_cache` filter here does not solve the problem in all cases, so I'm closing this PR and will prepare a new one. Btw. the filter itself is not new, it's already present in 2.2.4. I'm not sure, if it should be mentioned in README, other hooks are not. But it should be at least mentioned in [Wiki](https://github.com/pluginkollektiv/cachify/wiki/en-Hooks), I'll take care for that (at least in EN version).

PR #145, opened 2017-08-29
Allow HDD cache to be flushed also when current user is not an admin.

chesio:
I think the `current_user_can( 'manage_options' )` check in `_user_can_delete()` method has been introduced by mistake, it effectively breaks HDD cache flushing for any non-admin user.

### DIFF
--- a/inc/cachify_hdd.class.php
+++ b/inc/cachify_hdd.class.php
@@ -363,10 +363,6 @@ final class Cachify_HDD {
 			return false;
 		}
 
-		if ( ! current_user_can( 'manage_options' ) ) {
-			return false;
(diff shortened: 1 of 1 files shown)


bueltge:
Please add the change to the `CHANGELOG.md` so that users can read about this.

chesio:
@bueltge The issue, that this PR fixes, has been introduced quite recently (in 54ed868052e08d9df626f498380c9f644a8f6f34), so it's not something that should be reported to users. Rather the contrary, this PR reverts unintentional and undocumented breaking change (I believe it's been unintentional, but I requested review by @websupporter to be sure).

PR #148, opened 2017-12-04
Minor correction for compatibility with PHP 5.2

stklcode:
Passing method call to empty() is not available in PHP prior to 5.5, so add temporary local variable to restore compatibility.

I know, 5.2 is damn old and there are plenty of reasons to go for >5.5, so we might also raise requiremed version to 5.5, but a single line does the job and it only affects one caching method...


bueltge:
I hope is not necessary for your environment ;)

Issue #102, opened 2017-05-29
Cachify does not inform user that cache has been cleared when flushing cache from frontend


chesio:
When user clicks on Cachify flush button in admin bar when viewing front-end page, cache is cleared and page reloads, but there is no indication of success or error, so to the user it seems as nothing has happened.

Possible solutions (sorted by implementation complexity):
1. Do not show flush button in admin bar, when viewing non-admin page.
1. Redirect to Dashboard page in admin and show a notice there using admin notices mechanism like it is currently done for admin pages.
1. AJAX-ify cache flushing action and show any notice within the current page.

The AJAX-based solution would be the best in my opinion. Cache flushing in [Autoptimize](https://github.com/futtta/autoptimize/) plugin works this way and it seems more UX friendly to me. Also, it would inherently fix #100.

krafit:
I really like the AJAX solution too. What do you think, @bueltge, @websupporter?

chesio:
Question: do you want to support IE9 users? I have an implementation of AJAX-ified cache flushing ready, but it uses [CSS Animations](http://caniuse.com/#feat=css-animation) to rotate a spinner that indicates that flushing is in progress (I got inspired by cache flushing indicator from TYPO3).

I'm currently testing this feature on one of my sites, but I like it already: cache flushing without page reload is way faster.

websupporter:
> What do you think, [...], @websupporter?
sorry, totally missed that. Sounds good 😄

angcl:
#257 This pull request may be the answer. It's a version that registers an REST API endpoint which is used to clear the cache.

PR #150, opened 2018-02-05
Add "Clear Cache" Button 

00travelgirl00:
PR includes just a mockup, not a  functionality!


00travelgirl00:
Screenshot of the changes 

bildschirmfoto 2018-02-05 um 22 58 52

inc/cachify.settings.php
```diff
@@ -112,3 +112,19 @@
...
+<form method="post" action="options.php">
+	<?php settings_fields( 'cachify' ) ?>
+    <table class="form-table">
+        <tr>
+            <th scope="row">
+				<?php esc_html_e( 'Delete Caching', 'cachify' ) ?>
+            </th>
+            <td>
+                <label for="delete-cache">
+                    <button type="button" class="button-secondary">Clear Cache</button>
```
websupporter: esc_html_e( 'Clear Cache', 'cachify' );

websupporter:
Thanks @00travelgirl00,
so, we would basically need to add the actual "cache-flush"-functionality, right?

If you don't feel comfortable to add it, I can do that. Other than that, why is this "option" not together with the other options? Maybe because its not an option, but to me it seems a bit lost down there 🤔

krafit:
I'm not entirely sure, if this is fixing an actual problem 🤔 
I don't see an immediate benefit of adding another option and would rather add a brief description text alongside the *Cache expiration* option.

00travelgirl00:
@krafit I understand the "Cache expiration" as a automatic time period when the cache should be renewed. 

The Use case for the "Clear Cache" Button is, if I made a change but can't see it on the website. The first thing I want to do is to make sure its not the cache, so I would like to clear it to check, if the setting/code what ever is wrong or not. 

@websupporter I put it below, because its not a actual setting, that I could save. It is an other functionality, so I thought its better, to put it on an extra space. 

I made different mockups, also the "Clear Cache" Button next to the Save Changes Button, but I tought it could might be confusing for non experienced users. 

But we could place it above the links.

websupporter:
@krafit 
right now, we only have the trash icon on the top right. This might not be super obvious for new users. My first guess would also be to go to some settings page to flush the cache, which is something else than the cache expiration.

@00travelgirl00 
maybe we could just add a "flush cache now"-button underneath the hours-input field of the cache expiration? Around here: https://github.com/00travelgirl00/cachify/blob/e9cb0376b96497494326c7eab2e9c156716d73ae/inc/cachify.settings.php#L37-L38

stklcode:
> right now, we only have the trash icon on the top right

Wow, using the plugin for years, knowing the code and still never noticed that button as part of Cachify (or maybe forgot about it). So much for "not super obvous"... :open_mouth: Maybe move it to the dashboard widget next to the cache size thinking

I'd personally like to go even further and rework the call to `flush_total_cache(true)` call in die `validate_options()` method in favor of the proposed new button. (not point of this PR, but should be taken into count)
It's rather annoying that increasing the caching timeout or changing e.g. comment settings flushes all pre-existing values...

Flushing on backend change, drecrease of timeout and update of the blacklist is obviously still a good idea. And of course, making no-change save to flush works, but is not intuitive

If you know it's flushing on save OK, but on the UX perspective not perfect though. And yes, I'm with @00travelgirl00 to separate the button into the settings form, as it is basically no setting.

chesio:
Related: #102

I think the button (no matter where it ends up) would benefit from AJAX-ification.

In https://github.com/chesio/cachify/tree/issue-102, I have an AJAX solution for the not-super-obvious button in admin bar :-)

2ndkauboy:
@websupporter @krafit looks good to me, ready to merge.

PR #154, opened 2018-04-13
Patch 2

Syberspace:
Currently there is no way to decide against caching a given page after the `template_redirect` action has fired.  
This pull adds a filter `cachify_store_item` which fires before the actual `store_item` function of the selected caching method is called.  

This would allow developers to decide against caching a page at any point in the rendering process.  
The simplest way to achieve this would be adding `add_filter( 'cachify_store_item', '__return_false' );` anywhere in the codepath.

Examples for this would be HDD-cache in conjunction with a page that uses a nonce. The nonce would break after 24 hours if the cache isn't manually flushed.

Another example could be a page with dynamic sections that should not be cached but are not yet available at the time `template_redirect` fires.

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1302,14 +1302,13 @@ final class Cachify {
 	 * @return  string        Content of the page.
 	 */
 	public static function set_cache( $data ) {
-
 		/* Empty? */
 		if ( empty( $data ) ) {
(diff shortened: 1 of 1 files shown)


bueltge:
Short and helpful, also with style of the WP codex. Thanks.

Issue #151, opened 2018-02-05
Add "Clear Cache" functionality for a better ux


00travelgirl00:
While using cachify I'm missing the "clear cache" functionality, because for unexperienced users it is not quiet clear how to clear the caching and to check their changes. 

I have created a patch and a mockup, which only includes the design, not the functionality! 

PR: #150 

<img width="1232" alt="bildschirmfoto 2018-02-05 um 22 58 52" src="https://user-images.githubusercontent.com/21318377/35831374-b26beb42-0ac9-11e8-9ce8-3fa586c3d805.png">

te-online:
While there is the trash can in the admin bar to empty the cache, the icon is neither very descriptive nor does it seem available for non-admin users.

I'd rather add a label to the trash-can and/or replace the icon in the admin bar and additionally make it configurable to make it available to non-admin users as well.

00travelgirl00:
I not even noticed the trash can and did not know, its relate to the caching plugin. I think its important that you have the option to clear the cache also in the settings page of cachify, because thats the place where new users would look at. 

Maybe we can add the trash can symbol next to expiration with the label "clear cache" and the description "you can also use the trash can symbol in the admin bar". 

In addition we could also label the trash can in the admin bar, to be more clear.

te-online:
Yeah, it's a huge usability problem. I think your suggestions are good, and I'll see if I can get a pull-request done in the upcoming weeks.

websupporter:
@te-online,
thanks for having a look into this, please keep also the discussion on the PR #150 in mind, when drilling into this problem. :heart:

Zodiac1978:
Related: https://gist.github.com/glueckpress/0ccc84e6f93b89dee1d3

dionysous:
PR #150 is now merged. I relocated the flush button right under the input field in the **cache expiration** section. The buttons is colored orange.

When I used to visit the Cachify plugin settings page, I was always looking for a **flush button**. That's why I worked on @00travelgirl00 PR.

<img width="1018" alt="screenshot: cachify settings page with orange flush button" src="https://user-images.githubusercontent.com/12687777/45818931-656c6200-bce3-11e8-965c-000156d3de60.png">

te-online:
@dionysous This looks awesome! 😊  (I haven't had the time to do anything regarding the caption in the admin bar, though... 😳 )

te-online:
[...]

PR #158, opened 2018-09-20
Add description text on how to flush the cache manually

dionysous:
As mentioned in PR https://github.com/pluginkollektiv/cachify/pull/150#issuecomment-363343499 by @krafit, it would be a good way to add a description text in the **Cache expiration** section under the input. I created this PR as suggestion, it would look like this (in database and in HDD cache mode):

**Database cache mode:**
description-database-cache
**HDD cache mode:**
description-harddisk-cache

See: #151

### DIFF
--- a/inc/cachify.settings.php
+++ b/inc/cachify.settings.php
@@ -34,8 +34,8 @@ defined( 'ABSPATH' ) || exit;
 				
 				<p class="description">
 					<?php printf(
-						/* translators: Placeholder is the icon itself as dashicon */
-						esc_html__( 'Flush the cache by clicking the button below or the %1$s icon in the admin bar.', 'rh-50' ),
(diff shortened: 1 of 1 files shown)


dionysous:
**Note:** The gab between the orange button and the description text above is not that huge in reality. It was caused by an extra `<br />` element, which is not part of this PR.

bueltge:
Hi @dionysous, why do you use the translation domain `'rh-50'` in the new text? maybe a mistake? It should be `cachify`.

dionysous:
@bueltge It's `'cachify'` now >  https://github.com/pluginkollektiv/cachify/pull/158/commits/a0a0b6325fee3c587d0b46a76b729874ed010c2c  👍

PR #159, opened 2018-10-07
Admin bar icon – Add option to show text caption next to flush cache icon.

te-online:
Addresses UX problem described in #151.

### Behavior
- By default flush cache button in admin bar looks the same as always
- There is a new option right above the new flush cache button on the settings page that enables  the caption (for admins who fear their users/co-workers will miss the button)
- When the setting is true a text caption will be shown next to the icon inside the button within the admin bar

**Note:** If you have any suggestions regarding functionality, wording or code style, I'd be happy to discuss. This is my first PR in this repository.

### Screenshots
bildschirmfoto 2018-10-07 um 19 55 24
bildschirmfoto 2018-10-07 um 19 55 16

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -499,6 +499,7 @@ final class Cachify {
 				'use_apc'		 	=> self::METHOD_DB,
 				'reset_on_comment'  => 0,
 				'sig_detail'        => 0,
+				'show_adminbar_button_caption' => false
 			)
 		);
 	}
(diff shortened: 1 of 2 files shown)


bueltge:
Thanks a lot @te-online for your PR, I can't find hints or problems, also right code style.

krafit:
As @bueltge already mentioned: nice code style, awesome first PR - would love to see more like this in future 🙃 ❤️ Thank you @te-online.

I do however want to avoid introducing an option for a non-feature like a button caption. If we decide to add a caption to the button (and with the previous discussions in mind, I'm pretty convinced, we should), I should be displayed all the time, without an option to toggle it.

If you want to, you can change your PR accordingly. I'd be happy to merge it afterwards 🙂

te-online:
@bueltge @krafit Thanks for the kind words! 😊 I removed the option, going for this *decision* 😉 

I'm not sure about the addition of the translator annotation in form of the _x command on the settings page. I didn't revert that. Feel free to change how it suits you best before merging.

Thanks!

Zodiac1978:
Hi @te-online !

> I'm not sure about the addition of the translator annotation in form of the _x command on the settings page.

This is not an annotation, that's a misunderstanding. This is a *context*. The difference is:
A annotation (or better [translation comment](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#descriptions) is used to inform the translator about something. For example the meaning of the placeholder.

Context is used if the string has two meanings and needs disambiguation. For example: "Post" (verb, to post) and "Post" (noun, post vs. page).

Please see: https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#disambiguation-by-context

So please don't use the context in this case as this is not the right use case for it.

Thank you very much!

te-online:
@Zodiac1978 Okay, thanks for the heads up. I just learned about the translation comments, didn't know it was an officially supported thing 😳 Will adjust the PR later today! 💪

te-online:
@bueltge @Zodiac1978 Changes have been made as discussed. Do you want the commits to be squashed or can you select that option when merging?

bueltge:
The commits to this PR is fine, no separation is necessary. I think @krafit will merge this PR.

te-online:
@krafit Thanks a lot 😊That's so nice to hear!

Issue #167, opened 2019-02-25
Flush Cache automatically when creating/modifying/deleting Terms/Menus


derweili:
Currently Cachify auto-flushes the cache on these events:

- After publishing new posts
- After publishing new pages
- After publishing new custom post types
- After publishing new sheduled posts (only Cachify DB)
- After updating WordPress
- If you confirm the trash button on the adminbar

Is there a reason why the cache is not flushed for the following events?
- Menus
  - Creating Menus
  - Editing Menus
  - Deleting Menus
- Terms
  - Creating Terms
  - Editing Terms
  - Deleting Terms

I often have the case users edit Terms or Menus without clicking the Flush Cache Button. (depending on their user role, the Flush Cache Button is hidden so the can't click it)
I know I can extend the Plugin with another plugin but wouldn't it be great if this was a feature inside Cachify?

krafit:
Hi @derweili and thank you for your idea. 
You are right, I also would expect to see the cache flushed for all the events you listed. 

I'm not aware of a reason this is not the case right now, but that shouldn't prevent us from implementing it now.

I'd love to see this in v.2.4 and if you'd like to tinker around with it, feel free do to so, your contribution would be highly appreciated ❤️

PR #169, opened 2019-03-27
Flush cache on term change

derweili:
This PR is a fix for #167 

To create this functionality I hooked the `Cachify::flush_total_cache()` function on to three additional actions:

- `create_term` to clear the cache when a new term is created
- `edit_terms` to clear the cache when a term is updated.
- `delete_term` to clear the cache when a term is deleted

Because the nav menus are based on terms this modification also flushes the Cache when a nav menu is created, updated or deleted

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -91,7 +91,7 @@ final class Cachify {
 		);
 
 		/* Flush Hooks */
-		self::register_flush_cache_hooks();
+		add_action( 'init', array( __CLASS__, 'register_flush_cache_hooks' ), 10, 0 );
 
 		add_action(
 			'cachify_remove_post_cache',
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1098,63 +1098,30 @@ final class Cachify {
 
 		/* Define all default flush cache hooks */
 		$flush_cache_hooks = array(
-			array(
-				'action' => 'cachify_flush_cache',
-				'priority' => 10
-			),
-			array(
-				'action' => '_core_updated_successfully',
-				'priority' => 10
-			),
-			array(
-				'action' => 'switch_theme',
-				'priority' => 10
-			),
-			array(
-				'action' => 'before_delete_post',
-				'priority' => 10
-			),
-			array(
-				'action' => 'wp_trash_post',
-				'priority' => 10
-			),
-			array(
-				'action' => 'create_term',
-				'priority' => 10
-			),
-			array(
-				'action' => 'delete_term',
-				'priority' => 10
-			),
-			array(
-				'action' => 'edit_terms',
-				'priority' => 10
-			),
-			array(
-				'action' => 'user_register',
-				'priority' => 10
-			),
-			array(
-				'action' => 'edit_user_profile_update',
-				'priority' => 10
-			),
-			array(
-				'action' => 'delete_user',
-				'priority' => 10
-			)
+			'cachify_flush_cache' 				=> 10,
+			'_core_updated_successfully'	=> 10,
+			'switch_theme'								=> 10,
+			'before_delete_post'					=> 10,
+			'wp_trash_post'								=> 10,
+			'create_term'									=> 10,
+			'delete_term'									=> 10,
+			'edit_terms'									=> 10,
+			'user_register'								=> 10,
+			'edit_user_profile_update'		=> 10,
+			'delete_user'									=> 10
 		);
 
 		$flush_cache_hooks = apply_filters( 'cachify_flush_cache_hooks', $flush_cache_hooks );
 
 		/* Loop all hooks and register actions */
-		foreach ($flush_cache_hooks as $hook) {
+		foreach ($flush_cache_hooks as $hook => $priority) {
 			add_action(
-				$hook["action"],
+				$hook,
 				array(
 					'Cachify',
 					'flush_total_cache',
 				),
-				$hook["priority"],
+				$priority,
 				0
 			);
 		}
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1129,6 +1129,18 @@ final class Cachify {
 			array(
 				'action' => 'edit_terms',
 				'priority' => 10
+			),
+			array(
+				'action' => 'user_register',
+				'priority' => 10
+			),
+			array(
+				'action' => 'edit_user_profile_update',
+				'priority' => 10
+			),
+			array(
+				'action' => 'delete_user',
+				'priority' => 10
 			)
 		);
 
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1094,27 +1094,27 @@ final class Cachify {
 	 * @param   void
 	 * @return  void
 	 */
-	public static function register_flush_cache_hooks(){
+	public static function register_flush_cache_hooks() {
 
 		/* Define all default flush cache hooks */
 		$flush_cache_hooks = array(
-			'cachify_flush_cache' 				=> 10,
-			'_core_updated_successfully'	=> 10,
-			'switch_theme'								=> 10,
-			'before_delete_post'					=> 10,
-			'wp_trash_post'								=> 10,
-			'create_term'									=> 10,
-			'delete_term'									=> 10,
-			'edit_terms'									=> 10,
-			'user_register'								=> 10,
-			'edit_user_profile_update'		=> 10,
-			'delete_user'									=> 10
+			'cachify_flush_cache' => 10,
+			'_core_updated_successfully' => 10,
+			'switch_theme' => 10,
+			'before_delete_post' => 10,
+			'wp_trash_post' => 10,
+			'create_term' => 10,
+			'delete_term' => 10,
+			'edit_terms' => 10,
+			'user_register' => 10,
+			'edit_user_profile_update' => 10,
+			'delete_user' => 10
 		);
 
 		$flush_cache_hooks = apply_filters( 'cachify_flush_cache_hooks', $flush_cache_hooks );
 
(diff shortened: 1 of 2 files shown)


krafit:
Wow, thank you @derweili ❤️

@chesio could you do the honors and review this commit?

bueltge:
Maybe we should use a separate method that loop about all hooks there flush the cache. It should include a custom hook, so that uses have the possibility to remove hooks for flushing.
Currently we have really often to fire add_action and log of hooks.

derweili:
@bueltge thank's for your feedback. I think that makes sense.

I moved all the hooks to a new method `Cachify::register_flush_cache_hooks()`

I included a filter `cachify_flush_cache_hooks` so users can add or remove their hooks that should flush the cache.

bueltge:
Thanks a lot @derweili 
There was only an idea, not really relevant in the context of your PR, but is really great that you add this effort! :heart:

derweili:
I just found out the cache is not flushed when a user is created/updated/deleted.

Because the change of user data (e.g. display name) can cause changes on all posts, the cache should also flush on these event.

So I just added three further hooks to flush the total cache:

- `user_register`
- `edit_user_profile_update`
- `delete_user`

inc/cachify.class.php
```diff
@@ -1124,6 +1086,80 @@ private static function _is_logged_in() {
...
+				'action' => 'delete_user',
+				'priority' => 10
+			)
+		);
+
+		$flush_cache_hooks = apply_filters( 'cachify_flush_cache_hooks', $flush_cache_hooks );
+
+		/* Loop all hooks and register actions */
+		foreach ($flush_cache_hooks as $hook) {
+			add_action(
```
chesio: Please, add 4-th parameter (`$accepted_args`) to [add_action](https://developer.wordpress.org/reference/functions/add_action/) and set it to `0`. `Cachify::flush_total_cache()` accepts one parameter (`$clear_all_methods`), but I'm sure you don't want to have it set in this situation and it is unrelated to whatever the hooked actions pass as first parameter (like `$term_id` in case of [create_term](https://developer.wordpress.org/reference/hooks/create_term/) action etc.)

inc/cachify.class.php
```diff
@@ -91,6 +91,8 @@ public function __construct() {
 		);
 
 		/* Flush Hooks */
+		self::register_flush_cache_hooks();
```
chesio: This method should be called later, I would say [init](https://developer.wordpress.org/reference/hooks/init/) hook should suffice. Right now it [gets executed](https://github.com/pluginkollektiv/cachify/blob/master/cachify.php#L46-L52) in context of [plugins_loaded](https://developer.wordpress.org/reference/hooks/plugins_loaded/) hook, but this is too early for themes to hook to the new `cachify_flush_cache_hooks` filter.

So this would be right way to go:
```php
add_action( 'init', array( __CLASS__, 'register_flush_cache_hooks' ), 10, 0 );
```

Also, you need to make the method public, otherwise the callback will fail to execute.

inc/cachify.class.php
```diff
@@ -1124,6 +1086,80 @@ private static function _is_logged_in() {
 		return false;
 	}
 
+	/**
+	 * Register all hooks to flush the total cache
+	 *
+	 * @since   2.4.0  Make the function public
```
chesio: I'm not sure what you mean by that comment? It sounds more like `@todo` than `@since`.

inc/cachify.class.php
```diff
@@ -1124,6 +1086,80 @@ private static function _is_logged_in() {
...
+	 *
+	 * @param   void
+	 * @return  void
+	 */
+	private static function register_flush_cache_hooks(){
```
chesio: Please, make the method `public`. Although it is private by logic, it must be public when executed from within a hook (see also my other comment).

inc/cachify.class.php
```diff
@@ -1124,6 +1086,80 @@ private static function _is_logged_in() {
...
+
+		/* Define all default flush cache hooks */
+		$flush_cache_hooks = array(
```
chesio: One more thing: I would perhaps suggest to refactor `$flush_cache_hooks` array in `action` => `priority` fashion:
```php
$flush_cache_hooks = array(
    'cachify_flush_cache' => 10,
    '_core_updated_successfully' => 10,
    'switch_theme' => 10,
    ...
);
```

This way it is much easier to alter/remove hooks from the array when necessary, no for-loop or array_map/array_filter is necessary:
```php
add_filter( 'cachify_flush_cache_hooks', function ( $hooks ) {
    // Do not flush when post is trashed.
    unset( $hooks[ 'wp_trash_post' ] );
    // Change priority, flush later (for whatever reason).
    $hooks[ 'user_register' ] = 20;

    return $hooks;
});
```

inc/cachify.class.php
bueltge: Maybe `plugins_loaded` is a better option, runs earlier, runs before load translations, there write inside cache?

inc/cachify.class.php
bueltge: Really good point!

inc/cachify.class.php
chesio: Let me explain with example. Let's say someone wants to not flush the cache when a post is trashed. He adds the following code in his theme's `functions.php`:

```php
add_filter( 'cachify_flush_cache_hooks', function ( $hooks ) {
    unset( $hooks[ 'wp_trash_post' ] ); // Note: assumes that $flush_cache_hooks is refactored as I suggested
    return $hooks;
} );
```

However, `Cachify::register_flush_cache_hooks()` is executed in `plugins_loaded` action, so the call to `apply_filters( 'cachify_flush_cache_hooks', ... );` happens before `functions.php` is even loaded and the filter above is never run.

That's why `Cachify::register_flush_cache_hooks()` must be postponed to `init` action (or [`after_setup_theme`](https://developer.wordpress.org/reference/hooks/after_setup_theme/) at earliest).

inc/cachify.class.php
derweili: changed in f97b8157985ed0ff1125dac6cfba40a6bfd436e0

inc/cachify.class.php
derweili: @krafit  added the related issue to milestore 2.4
That's why I added @since 2.4.0

What would you recommend?

inc/cachify.class.php
derweili: updated the PR and included the recommended changes

inc/cachify.class.php
bueltge: `@since` should store only the version or date since we have this method, var. A change should not a part of the doc-bloc, especially not the `@since` tag.

bueltge:
Hi @derweili please leave also this changes inside the Changelog, so that we have a small list of features, changes inside this file.

derweili:
@bueltge I updated the changelog

inc/cachify.class.php
derweili: So should I remove the `@since` tag completely or only the description?

chesio: Only the description.

chesio:
@derweili, thanks for the fixes/updates. I'll review them later today!

inc/cachify.class.php
derweili: @chesio  ok, removed it.

@bueltge don't know. I think the other open tasks on the 2.3.0 milestone are  much more complex than this PR. So including this PR into 2.3.0 should not delay the 2.3.0 release.

krafit: @bueltge I added this to the `2.4.0` milestone because I did not expect this task to be fixed so quickly. But since @derweili did a stellar job here, we should move this to the `2.3.0` milestone and publish it as soon as possible.

bueltge:
Thanks @derweili For me all is fine, great PR.
Now we should wait that @chesio finally this PR.

inc/cachify.class.php
```diff
@@ -1124,6 +1086,48 @@ private static function _is_logged_in() {
...
 
+	/**
+	 * Register all hooks to flush the total cache
+	 *
+	 * @since   2.4.0
+	 *
+	 * @param   void
+	 * @return  void
+	 */
+	public static function register_flush_cache_hooks(){
```
chesio: Code style: please, add space before `{`.

inc/cachify.class.php
```diff
@@ -1124,6 +1086,48 @@ private static function _is_logged_in() {
...
+			'edit_terms'									=> 10,
+			'user_register'								=> 10,
+			'edit_user_profile_update'		=> 10,
+			'delete_user'									=> 10
+		);
+
+		$flush_cache_hooks = apply_filters( 'cachify_flush_cache_hooks', $flush_cache_hooks );
+
+		/* Loop all hooks and register actions */
+		foreach ($flush_cache_hooks as $hook => $priority) {
```
chesio: Code style: please, add space after `(` and before `)`.

inc/cachify.class.php
```diff
@@ -1124,6 +1086,48 @@ private static function _is_logged_in() {
...
+	 * @since   2.4.0
+	 *
+	 * @param   void
+	 * @return  void
+	 */
+	public static function register_flush_cache_hooks(){
+
+		/* Define all default flush cache hooks */
+		$flush_cache_hooks = array(
+			'cachify_flush_cache' 				=> 10,
```
chesio: Nitpicking: please, fix the white-space formatting here: either align `=>` on every line with proper number of tabs or use single space only before and after every `=>`.

chesio:
Thanks @derweili, your PR looks good to me now!

bueltge:
So @krafit we should merge and add to a version tag. Thanks to all of you!

Issue #161, opened 2018-10-21
Check for Gutenberg compatibility


krafit:
WordPress 5.0 will be here soon. Time for us to check, if our plugins are compatible with the Gutenberg Editor and WP 5.0 (a beta version is yet to be published). 

- [x] check (Gutenberg) editor with this plugin
- [x] check plugin settings site with active Gutenberg
- [x] fix potential issues
- [x] add `Tested up to: 5.0` to the plugins readme
- [x] publish changes as minor update or with upcoming plugin update

florianbrinkmann:
I tested saving of options, and if the minimizing option works (it does). I don’t know how to test if the cache is actually created/stored/updated …

Zodiac1978:
`post_submitbox_misc_actions` is not fired anymore. 
https://github.com/WordPress/gutenberg/issues/12764

Therefore our additional settings are not displayed anymore:
https://github.com/pluginkollektiv/cachify/wiki/en-Cache-Methods#control-of-cache-empties-for-article-updates

stklcode:
What about adding this to the global settings like "Default action after saving a post" (clear post, clear all, clear nothing)?
In addition adding a "clear cache or this page" functionality to a submenu for the admin-bar icon (which currently clears the site cache) would also be nice.

I think many users have never used the settings next to the editor anyway (I never did), so in my eyes the above changes would make the clearing functionality more intuitive.

PR #176, opened 2019-10-06
Configure action on modified posts in global settings

stklcode:
Related to https://github.com/pluginkollektiv/cachify/issues/161#issuecomment-519182315

The select box to either flush the total cache or just remove the current post, which had been present in the sidebar of the classic editor is no longer available with Gutenberg.

I've migrated this into a global settings flag, defaults to flushing the site cache.
To keep it consistent with legacy sites, I also completely removed the submitbox extension.

So it is now possible to not flush the total cache after every post edit on Gutenberg sites. However it is no longer possible to quickly override the setting for a single action.
From my point of view is is now more clear how the cache behaves, because it is always the same.


krafit:
I really like the approach, thanks @stklcode 🤗

PR #179, opened 2019-11-11
 #162 added description for option "Flush the cache at new comments"

### LINKED ISSUE #162: Adding a comment via the backend does not respect comment setting
If the checkbox for clearing the cache when posting a comment is **not** checked, the cache is emptied anyways if a comment is added via the backend (I tested it in the comments view, answering to another comment). Works like expected (not clearing the cache after a new comment) if the comment is added via the frontend.

00travelgirl00:
closes #162

### DIFF
--- a/inc/cachify.settings.php
+++ b/inc/cachify.settings.php
@@ -70,6 +70,7 @@ defined( 'ABSPATH' ) || exit;
 					<label for="cachify_reset_on_comment">
 						<input type="checkbox" name="cachify[reset_on_comment]" id="cachify_reset_on_comment" value="1" <?php checked( '1', $options['reset_on_comment'] ); ?> />
 						<?php esc_html_e( 'Flush the cache at new comments', 'cachify' ); ?>
+						<?php esc_html_e( '(Only works for comments which comes from the front-end, not added from the back-end)', 'cachify' ); ?>
 					</label>
 					<p class="description">
 						<?php esc_html_e( 'If selected, the site cache will be flushed, otherwise only the corresponding post is removed from the cache.', 'cachify' ); ?>


inc/cachify.settings.php
```diff
@@ -70,6 +70,7 @@
 					<label for="cachify_reset_on_comment">
 						<input type="checkbox" name="cachify[reset_on_comment]" id="cachify_reset_on_comment" value="1" <?php checked( '1', $options['reset_on_comment'] ); ?> />
 						<?php esc_html_e( 'Flush the cache at new comments', 'cachify' ); ?>
+						<?php esc_html_e( '(Only works for comments which comes from the front-end, not added from the back-end)', 'cachify' ); ?>
```
bueltge: The singular verb comes does not appear to agree the plural subject comments, I mean it should use come.

bueltge (changes requested):

inc/cachify.settings.php
bueltge: Small mindset. Maybe we should also add a point after each sentence. However currently we it not always on the messages.

derweili:
As far as I can check this functionality, when adding comments from the frontend, backend or REST-API there is no difference in the "Clear Cache Behavior".

In both cases, if the checkbox is checked or not, the frontend and backend work the same for me.

florianbrinkmann:
I misunderstood the option when opening the issue that this PR is for, it indeed works like it should.

florianbrinkmann:
PS: thanks for opening a PR to fix the issue, @00travelgirl00, and sorry for the work resulting from my misunderstanding.

PR #181, opened 2019-11-20
replace TEMPLATEPATH with get_template_directory()

### LINKED ISSUE #180: Use of undefined constant TEMPLATEPATH
When I enable cachify on my webpage I get the warning
```Warning: Use of undefined constant TEMPLATEPATH - assumed 'TEMPLATEPATH' (this will throw an Error in a future version of PHP) in .../wp-content/plugins/cachify/inc/cachify.class.php on line 1123```
Is there a fix for this? If not I would try and fix it myself and create a pr.

### PR DESCRIPTION
**Pull request by stklcode**
closes #180

### DIFF
--- a/inc/cachify.class.php
+++ b/inc/cachify.class.php
@@ -1043,12 +1043,13 @@ final class Cachify {
 	 * Check for mobile devices
 	 *
 	 * @since   0.9.1
-	 * @change  2.2.2
+	 * @change  2.3.0
 	 *
 	 * @return  boolean  TRUE if mobile
 	 */
 	private static function _is_mobile() {
-		return ( strpos( TEMPLATEPATH, 'wptouch' ) || strpos( TEMPLATEPATH, 'carrington' ) || strpos( TEMPLATEPATH, 'jetpack' ) || strpos( TEMPLATEPATH, 'handheld' ) );
+		$templatedir = get_template_directory();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: inc/cachify.class.php [hunk 1]
```diff
@@ -1043,12 +1043,13 @@ private static function _is_index() {
...
 	 *
 	 * @since   0.9.1
-	 * @change  2.2.2
+	 * @change  2.3.0
 	 *
 	 * @return  boolean  TRUE if mobile
 	 */
 	private static function _is_mobile() {
-		return ( strpos( TEMPLATEPATH, 'wptouch' ) || strpos( TEMPLATEPATH, 'carrington' ) || strpos( TEMPLATEPATH, 'jetpack' ) || strpos( TEMPLATEPATH, 'handheld' ) );
+		$templatedir = get_template_directory();
```
bueltge: If you set a variable for this, maybe we should unset this var to reduce the storage. However, if users use a storage cache there are faster if you always use the function, not a variable. But I write about so small seconds, so that it is also okay for me in this case.

stklcode: `get_template_director()` has at least 5 levels of nested method calls, 1 option to load and triggers 3 filter hooks, so it **might** perform better. But we can't tell what is actually called and how well OPcache performs here.

Anyway I didn't do any tests here and simply ignored any potential overheads.

bueltge:
Yes, that it what I mean, only a hint, we should not create a over-engineering in this case.

PR #190, opened 2020-04-17
use do_robots

### LINKED ISSUE #185: Automate robots.txt change?
In our wiki we tell our users this:

> To ensure that Google and other search engines don't index the static contents of the cache folder (otherwise duplicate content), the robots.txt file in the main directory of a WordPress installation should be expanded by disallow the path to the cache file.

But WordPress is using a "virtual" generated robots.txt, so the user will not find the file to extend it. He/she needs to copy the content from the virtual one and create a new one. 

See: 
https://github.com/pluginkollektiv/cachify/wiki/en-Cache-Methods#robotstxt-adjustion
https://github.com/pluginkollektiv/cachify/wiki/de-Caching-Methoden#anpassung-der-robotstxt

WordPress has this virtual robots.txt since version 2.1 and there is a filter we can use to add us to this file:
https://github.com/WordPress/WordPress/blob/d5f942d7bae2048a54d16fb320b34c841b1d7777/wp-includes/functions.php#L1655

So I would recommend we do this automatically with the filter and do not ask the user to do this.

I have checked other cahing plugins for the `robots_txt` filter, but no one does this. So I am not sure if this is needed altogether.

timse201:
resolves: #185


inc/cachify.class.php
```diff
@@ -473,28 +470,14 @@ private static function _get_options() {
...
 	 *
-	 * @param   string $data  Original content of dynamic robots.txt.
-	 * @return  string        Modified content of robots.txt.
 	 */
-	public static function robots_txt( $data ) {
+	public static function robots_txt() {
 		/* HDD only */
-		if ( self::METHOD_HDD !== self::$options['use_apc'] ) {
-			return $data;
+		if ( self::METHOD_HDD === $options['use_apc'] ) {
```
stklcode, line 478: `$options` is not defined in this scope. Should use `self::$options` reference.

inc/cachify.class.php
```diff
@@ -247,11 +247,8 @@ public function __construct() {
 			);
 
 			add_action(
-				'robots_txt',
-				array(
-					__CLASS__,
-					'robots_txt',
-				)
+				'do_robots',
+				'robots_txt'
```
stklcode, line 251: There is no public method `robots_txt`. Why has the reference to the static method been changed?
```php
array(
	__CLASS__,
	'robots_txt',
)

**Review by stklcode — CHANGES_REQUESTED** [review]

File: inc/cachify.class.php [hunk 3]
```diff
@@ -473,28 +473,14 @@ private static function _get_options() {
...
+		if ( self::METHOD_HDD === self::$options['use_apc'] ) {
+			echo 'Disallow: */cache/cachify/';
 		}
-
-		/* Parse site URL */
-		$url_parts = wp_parse_url( site_url() );
-
-		/* Output */
-		$data .= sprintf(
-			'%2$sDisallow: %1$s/wp-content/cache/cachify/%2$s',
```
websupporter, line 492: Should this respect the `WP_CONTENT_DIR` constant?

stklcode: For the documentation we just changed to `*` (#194), but in this point we _could_ be more precise.

timse201: we can use
printf(
	'Disallow: %s/cache/cachify/',
	wp_make_link_relative( content_url() ),
);
if you want. But when you move wp-content to another location or rename it then you don't wan't to leak it's path to public.

florianbrinkmann:
@stklcode it looks like all changes that you requested in your review were made, is that correct (just asking because of the red warning »1 review requesting changes« above the merge button)?

bueltge (approved):

florianbrinkmann (approved):
LGTM

PR #194, opened 2020-04-19
updated links

timse201:


readme.txt
```diff
@@ -94,7 +94,7 @@ Compared to APC (Alternative PHP Cache), PHP OPCache is not able to contain cont
 To ensure that Google and other search engines do not index the static contents of the cache folder (otherwise there could be duplicate content), the robots.txt file which is located in the main directory of a WordPress installation should be expanded by disabling the path to the cache file (disallow ). This issue should only happen if you use a *static robots.txt* or you changed the *wp-content* location. And so might look a robots.txt:
 
 `User-agent: *
-Disallow: /wp-content/cache/cachify/
```
stklcode, line 97: The only non-formal change here.
Only necessary if `WP_CONTENT_FOLDERNAME ` is redefined or the path is rewritten in another way. If done so, one - hopefully - doesn’t need that hint. Only relevant if .htaccess is used and manually created anyway.

However I don’t see any real argument against the asterisk in this place.

stklcode (approved):
LGTM.

Left a short comment on the only non-trivial change.

PR #198, opened 2020-10-03
Make hours translatable

timse201:


inc/cachify.settings.php
```diff
@@ -25,7 +25,8 @@
 			</th>
 			<td>
 				<?php if ( self::METHOD_HDD === $options['use_apc'] ) : ?>
-					<input type="number" min="0" step="1" name="cachify[cache_expires]" id="cachify_cache_expires" value="" disabled="disabled" class="small-text" /> Hours
```
bueltge, line 28: @timse201 Have you test this change in the UI? Because of the missing space before hours would be a problem.
`class="small-text" /> Hours` <-- original

Thanks for the PR and feedback on my small comment.

bueltge (changes requested):

inc/cachify.settings.php
MatzeKitt: A newline should also reflect a space in an HTML document. 🙂

bueltge: Yes, I mean also too. But it would be fine we have a human test about the UI.

stklcode: Just checked manually. As expected, no differences in the UI :+1:

stklcode (approved):

PR #192, opened 2020-04-18
introduce WPCS for PHP

stklcode:
Build on top of #191 

First syntax-/comments-only changes applied, yet WIP (feel free to contribute)

PR #191, opened 2020-04-18
minify CSS using Composer

stklcode:
* minify CSS using Composer instead of Grunt
* introduce Stylelint checks (CI to be done along with PHPCS/PHPUnit changes WIP)
* remove minified resources from SCM
  * add Composer build step to GH action for automatic deployment

PR #202, opened 2021-01-15
Add code style check and build script

patrickrobrecht:
- Adds a code style check for CSS files using stylelint. Existing violations are resolved already.
- Adds a code style check for PHP files. Existing violations are fixed or allowed via `phpcs:ignore`.
- Defines a build script for creating the `.min.css` files. The currently versioned `.min.css` files are removed.
- Changes the GitHub Actions to include a `composer install`. In addition the versions of the steps are updated and `readme.txt` is used instead of `README.md`

EDIT: I just saw that #191 and #192 are already open for the same issues. So this PR is an alternative to the other open PRs.


.github/workflows/wordpress-plugin-asset-update.yml
```diff
@@ -8,10 +8,11 @@ jobs:
     name: Push to master
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
+    - uses: php-actions/composer@v1
```
stklcode: Any reason to not use latest v5 of the composer action?
(same for the deployment action)

inc/class-cachify.php
```diff
@@ -696,7 +701,7 @@ public static function process_flush_request( $data ) {
 		}
 
 		/* Check nonce */
-		if ( empty( $_GET['_wpnonce'] ) || ! wp_verify_nonce( $_GET['_wpnonce'], '_cachify__flush_nonce' ) ) {
+		if ( empty( $_GET['_wpnonce'] ) || ! wp_verify_nonce( sanitize_text_field( wp_unslash( $_GET['_wpnonce'] ) ), '_cachify__flush_nonce' ) ) {
```
stklcode: Nonce generation is pluggable behavior, so it _might_ contain characters that do not pass `sanitize_text_field()`.

patrickrobrecht: I changed that to `sanitize_key`. There is a long discussion on https://github.com/WordPress/WordPress-Coding-Standards/issues/414 and https://github.com/WordPress/WordPress-Coding-Standards/issues/869 whether WPCS should throw an error at all and whether `sanitize_key` should be used or not. Alternative would be to add another ignore rule for that.

.github/workflows/wordpress-plugin-asset-update.yml
patrickrobrecht: No, I've updated the workflows to use the latest version of the actions.

inc/class-cachify.php
stklcode: I am aware of the discussions and totally fine with `sanitize_key`.

stklcode (approved):
Added some minor changes
* .distignore ignored in git export
* WPCS stylelint config 17 -> 19

Good to merge imho.

PR #209, opened 2021-03-27
use printf

timse201:


stklcode:
Thanks for your PR.

I personally prefer the semantics of `echo` over ` print` when used to output strings without any control structures, but there are certainly arguments for both variants. We should keep it consistent across the project (it's not entirely consistent, yet).

Some points for discussion:
* `printf` returns a result (number of characters) that is ignored, while `echo sprintf` does not return any value
* `printf` is 6 characters less than `echo sprintf` (obvious)
* `printf` is _slightly_ faster than `echo sprintf` in common cases
  <details>
  <summary>PHPBench example</summary>

  | PHP    | echoSprintf1      | echoSprintf2      | echoSprintf3      | printf1           | printf2          | printf3           |
  | ------ | ----------------- | ----------------- | ----------------- | ----------------- | ---------------- | ----------------- |
  | 8.0.3  | 0.613μs (±1.205%) | 0.664μs (±2.68%)  | 0.705μs (±0.534%) | 0.600μs (±1.625%) | 0.651μs (±0.8%)  | 0.694μs (±2.16%)  |
  | 7.4.16 | 0.394μs (±0.334%) | 0.532μs (±0.186%) | 0.672μs (±0.077%) | 0.379μs (±1.278%) |0.531μs (±0.731%) | 0.671μs (±0.631%) |

  ```php
  <?php

  /**
   * @BeforeMethods({"init"})
   * @AfterMethods ({"close"})
   * @Revs(1000000)
   * @Iterations(10)
   */
  class EchoSprintfBench {
      public function init() {
          ob_start();
      }

      public function close() {
          ob_end_clean();
      }

      public function benchEchoSprintf1() {
          echo sprintf('number %s', 'one');
      }

      public function benchEchoSprintf2() {
          echo sprintf('number %s %s', 'one', 'two');
      }

      public function benchEchoSprintf3() {
          echo sprintf('number %s %s %s', 'one', 'two', 'three');
      }

      public function benchPrintf1() {
          printf('number %s', 'one');
      }

      public function benchPrintf2() {
          printf('number %s %s', 'one', 'two');
      }

      public function benchPrintf3() {
          printf('number %s %s %s', 'one', 'two', 'three');
      }
  }
  ```
  </details>
* we do use `printf` twice while all other outputs, including the 7 changed here, run through `echo`
  https://github.com/pluginkollektiv/cachify/blob/041d60ad0d675b4bf153d1db95c3faf250f37566/inc/class-cachify.php#L1606-L1609
  https://github.com/pluginkollektiv/cachify/blob/041d60ad0d675b4bf153d1db95c3faf250f37566/inc/cachify.settings.php#L42-L48

stklcode (approved):

PR #213, opened 2021-04-04
AddDefaultCharset to hdd htaccess setup page

timse201:


stklcode:
Should we really remove this block from the FAQ?

timse201:
Another suggestion - in memcached and hdd nginx, charset is present in example code. So we could include it there in the same way. So we don't need it in FAQ and only need a hint in the documentation.

krafit:
> Should we really remove this block from the FAQ?

No, we shouldn't. But I don't think there is any harm in having it in two places.

stklcode (approved):

PR #215, opened 2021-04-10
move to wp style for sample code

### LINKED ISSUE #32: General UI Optimizations

timse201:
WP is using textareas for code examples.

- no inline style
- now max. 16 lines are shown - a scrollbar is present if it is more than this and textarea is sizeable
- textarea is readonly, too
- white-space: pre-wrap is in wp style

closes: #32 
#39


inc/setup/cachify.apc.htaccess.php
```diff
@@ -29,8 +29,7 @@
 		</tr>
 	</table>
 
-	<div style="background:#fff;border:1px solid #ccc;padding:10px 20px">
-		<pre style="white-space: pre-wrap">
+	<textarea rows="5" class="large-text code" name="code" id="cachify-code" readonly>
```
stklcode: Is there a specific reason to specify an ID instead of an additional class?

timse201: readonly is forcing it to gray. so we have to use !important or id

stklcode: We have to use a selector that is more precise than `textarea[readonly]`. `.cachify-code` is not.

`#cachify-code` is such selector, but `textarea.cachify-code` also works.

stklcode (approved):

PR #217, opened 2021-04-11
introduce GH action to verify coding standards

stklcode:
There are no real tests yet, but we do enforce coding standards. Add a GitHub action for verification.


patrickrobrecht (approved):
LGTM.

PR #219, opened 2021-04-13
Load WP-CLI only if needed

timse201:


cachify.php
```diff
@@ -91,13 +82,16 @@
  * @param string $class the class name.
  */
 function cachify_autoload( $class ) {
-	if ( in_array( $class, array( 'Cachify', 'Cachify_APC', 'Cachify_DB', 'Cachify_HDD', 'Cachify_MEMCACHED', 'Cachify_CLI' ) ) ) {
-		require_once(
-			sprintf(
+	if ( in_array( $class, array( 'Cachify', 'Cachify_APC', 'Cachify_DB', 'Cachify_HDD', 'Cachify_MEMCACHED' ), true ) ) {
```
stklcode, line 85: What is the benefit of removing the `Cachify_CLI` class from the autoloader?

inc/class-cachify-cli.php
```diff
@@ -55,53 +55,50 @@ public static function get_cache_size( $args, $assoc_args ) {
...
-		WP_CLI::add_command(
-			'cachify flush',
+/**
+ * Register CLI Commands
+ *
+ * @since   2.3.0
+ * @change  2.3.0
+ */
+/* Add flush command */
+WP_CLI::add_command(
```
stklcode, line 67: This is a class file which should preferably define a class. With this change, the file has 2 purposes, define the class and execute code. I personally dislike that mixture, if there is no significant benefit.

Is there a point I don't see right now, why this way of calling should be preferred over `add_action( 'cli_init', array( 'Cachify_CLI', 'add_commands' ) );` ?

cachify.php
```diff
@@ -91,13 +82,16 @@
...
 				CACHIFY_DIR,
 				strtolower( str_replace( '_', '-', $class ) )
-			)
 		);
 	}
 }
+
+/* Load WP-CLI */
+if ( defined( 'WP_CLI' ) && WP_CLI && class_exists( 'WP_CLI' ) ) {
+    require_once CACHIFY_DIR . '/inc/class-cachify-cli.php';
```
stklcode, line 96: Why not simply move the `add_action( ... )` call to this line?

stklcode (changes requested):
Thanks for your PR.

The conditional is a good idea, however maybe I just don't get the point why the whole refactoring is necessary or beneficial to achieve this. This was surely not done for no reason, so would you please add some words of explanation for your solution?

timse201:
Seems like it's already directly implemented:
https://github.com/wp-cli/wp-cli/issues/4818

but documentation is not up to date:
https://github.com/wp-cli/handbook/issues/578

PR #220, opened 2021-04-13
ignore code tag - minify

timse201:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code


stklcode:
By definition there is no need to retain the format of an inline `<code>` block, just text-level semantic ([ref](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element)) compared to the `<pre>` tag which is a grouping block for pre-formatted content ([ref](https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element)). Most minifiers I've worked with will squash spaces and breaks by default. Most browsers also do ignore spaces and newlines, so no need to preserve them in the HTML code.

Our minification regex might not be perfect, but it does not match nested `<pre><code>...</code</pre>` tags, so we are fine if the code is wrapped.

Live examples for browser output:
`<code>test   me</code>`: (<-- enforced with NBSP symbols for demonstration)
<code>test   me</code> (<-- regular 3 spaces)

`<pre><code>test   me</code></pre>`:
<pre><code>test   me</code></pre>


**Question:**
Is there any WP standard output or popular plugin requiring pre-formatted inline `<code>`?

timse201:
yes this is the intended behavior because pre is on ignore list.
scenarios are posts, pages, comments (and custom themes)

< pre > tag 3 million (nerdydata)
< code > tag 8.5 million (nerdydata)
code search for < code > but has no < pre > tag (over 5.5 million sites):
https://www.nerdydata.com/reports/code-and-not-pre/0f9b5849-28f1-4926-b070-07954ffe046d

your test < pre >< code> 150,000:
https://www.nerdydata.com/reports/pre-code/0dd7b75e-cdbb-45a3-9328-3f0d8f68c2e6

stklcode:
OK, but why is this an argument not to minify standalone `<code>`? The semantics of `pre` and `code` are intentionally different.

Example: this comment text here contains 8 `<code>` tags and no `<pre>`, none of them must remain unminified. So what’s the point behind the numbers? Browser don’t render spaces and line breaks inside `<code>` by default, so why should we keep them?

I do use `<code>` quite often, too, but only for non-preformatted inline use that may be minified without any problems at all. If not, I do wrap in in an additional `<pre>`. And that’s what WP does for code-_blocks_, too. If any piece of code does not, I’d consider this to be a misinterpration of HTML tags.

However, if there is some widely used code out there that does misuse the tags, I’m willing to change my opinion and exclude it by default. Right now I don’t see any argument for that. (I won’t overrule the team though, just one opinion)

krafit:
As this PR has been untouched for quite some time and I'm not entirely convinced, we have to do something here. I'll close this for now. Please feel free to open an issue in the future if you still think this should be addressed.

PR #222, opened 2021-04-14
Update in_array() to use strict checks

timse201:
Update instances of in_array() to use strict checks to prevent odd results in some cases and add wp-cli comment till https://github.com/wp-cli/handbook/issues/578 is closed.


stklcode (approved):
LGTM

Fixed the argument indentiation due to dropped parenthesis after `require_once`.

PR #224, opened 2021-04-15
prevent unnecessary cache clearing

### LINKED ISSUE #223: Flush cache overwork
Right now the entire cache is cleared when
user_register - don't know why
edit_user_profile_update - why should it clear all and not it's own profile page cache
delete_user - makes sense if there is a link to the profile

and add some third parties:
_to clear entire cache:_
autoptimize_action_cachepurged
_when needed:_
woocommerce_product_set_stock
woocommerce_variation_set_stock
woocommerce_product_set_stock_status
woocommerce_variation_set_stock_status

timse201:
- flush cache when comment is in database and not before
- flush when comment is or was approved
- move comment hooks to comment 'flush hooks'
- rename functions for better code comprehension
- removes user_register and edit_user_profile_update from flush total cache

- because of edit_comment $data we need wp 4.6 or greater

#223


inc/class-cachify.php
```diff
@@ -807,15 +774,22 @@ public static function flush_notice() {
...
+	 * @param   array	$comment	Comment data
 	 */
-	public static function edit_comment( $id ) {
-		if ( self::$options['reset_on_comment'] ) {
-			self::flush_total_cache();
-		} else {
-			self::remove_page_cache_by_post_id(
-				get_comment( $id )->comment_post_ID
-			);
+	public static function comment_edit( $id, $comment ) {
```
stklcode: I wouldn't rename this method, as it's a breaking change.
Making the second parameter optional would also be nice, to not break the current API.

inc/class-cachify.php
```diff
@@ -825,21 +799,18 @@ public static function edit_comment( $id ) {
...
-	 * @param   array $comment   Array of properties.
-	 * @return  mixed            Comment status.
+	 * @param	integer			$id			Comment ID
+	 * @param	integer|string	$approved	Comment status
 	 */
-	public static function pre_comment( $approved, $comment ) {
+	public static function new_comment( $id, $approved ) {
 		/* Approved comment? */
-		if ( 1 === $approved ) {
+		if ( $approved === 1 ) {
```
stklcode: Yoda-condition I'd prefer here.

inc/class-cachify.php
```diff
@@ -852,8 +823,8 @@ public static function pre_comment( $approved, $comment ) {
 	 * @param   string $old_status  Old status.
 	 * @param   object $comment     The comment.
 	 */
-	public static function touch_comment( $new_status, $old_status, $comment ) {
-		if ( $new_status !== $old_status ) {
+	public static function comment_status( $new_status, $old_status, $comment ) {
```
stklcode, line 985: I like the new method name, however we should either introduce an alias and deprecate the old or keep the name for now.

inc/class-cachify.php
```diff
@@ -825,21 +799,18 @@ public static function edit_comment( $id ) {
...
 	 * @change  2.1.2
 	 *
-	 * @param   mixed $approved  Comment status.
-	 * @param   array $comment   Array of properties.
-	 * @return  mixed            Comment status.
+	 * @param	integer			$id			Comment ID
+	 * @param	integer|string	$approved	Comment status
 	 */
-	public static function pre_comment( $approved, $comment ) {
+	public static function new_comment( $id, $approved ) {
```
stklcode: There would be an easy way to keep the existing method without breaking the API I guess.

stklcode (changes requested):
Thanks for your work.

The changes in general are well suited for point release like 2.4, however they do introduce some breaking changes and I'm not sure whether we should raise the major version for this, if there is an viable way to add the new behavior non-breaking.

inc/class-cachify.php
timse201: https://developer.wordpress.org/reference/hooks/comment_post/
comment_post was introduced in 1.2

inc/class-cachify.php
timse201: only 1.7% of all wp users are left on wp 4.4 and 4.5
if i am right automatic plugin updates were only introduced since wp 5.5
but it's up to you

stklcode: It’s not only about WP, but about the code. Wouldn’t care too much about WP <4.6. But we have no idea who might be using this method, potentially called from any other plugin or some custom hook for whatever reason that might be.

timse201: In general. Sorry, but I don't know how to tell this. I don't want to blame you or feel blamed. How would you like to improve something without breaking or to modify **internal** things from time to time. Especially when it comes to the performance of a caching plugin.
If it's about code readability and you make everything twice - there will be no benefit for new coders. So there should be a renaming where it is needed for **one release** and leave it as is.

I edited it because it was the same function name as the WP function. There could be some confusion for new coders who want to read the code.

krafit: I think @stklcode has a point, we should avoid renaming this method.

stklcode: From my point of view: internal === ( private || annotated with `@internal` (PHPdoc) or `@private` (JSdoc) )

If there are good reasons to change things, we should change and raise the version accordingly. You, @timse201 are absolutely right, changes need changes and from time to time they are breaking and that’s fine.

Your reasons are good, so we might rename it, but to not mess up with the community, we should either
* keep and deprecate the old method or
* announce the change before the release

An unused, deprecated method does not do any harm here and gives other users/devs the chance to migrate before it gets removed.

> If it's about code readability

It’s about maintainability, too. Beautiful, perfectly readable code that changes unexpectedly is a pain to maintain for consuming projects.

To not forget about such duplications, we might just open a follow-up issue/PR to clean this up and assign it to a 3.0 milestone, so it’s visible.

krafit: Let's take the deprecation route and keep the old method around for now instead of simply renaming it.

inc/class-cachify.php
krafit: Yes, let's deprecate this one as well.

krafit (changes requested):

inc/class-cachify.php
```diff
@@ -1108,8 +1128,6 @@ public static function register_flush_cache_hooks() {
 			'create_term' => 10,
 			'delete_term' => 10,
 			'edit_terms' => 10,
-			'user_register' => 10,
-			'edit_user_profile_update' => 10,
```
florianbrinkmann: I would not remove those, like @stklcode wrote in https://github.com/pluginkollektiv/cachify/issues/223#issuecomment-820627103, user info can be displayed in many places.

inc/class-cachify.php
```diff
@@ -808,14 +775,33 @@ public static function flush_notice() {
...
+	/**
+	 * Remove page from cache or flush on comment edit.
+	 *
+	 * @since 2.4 Replacement for edit_comment($id) with additional comment parameter.
+	 *
+	 * @param integer $id      Comment ID.
+	 * @param array   $comment Comment data.
+	 */
+	public static function comment_edit( $id, $comment ) {
+		$approved = (int) $comment['comment_approved'];
```
florianbrinkmann, line 905: Couldn’t we just fetch the comment data with `get_comment()` instead of using the data param added in 4.6, so we do not need to increase the min WP version for that?

stklcode:
> because of edit_comment $data we need wp 4.6 or greater

We require 4.7 since #268, so that's no longer an issue here.

inc/class-cachify.php
stklcode: resolved in feab290d58cb32083c8fd2ae105930cab2dac7d3

inc/class-cachify.php
stklcode: resolved in feab290d58cb32083c8fd2ae105930cab2dac7d3

inc/class-cachify.php
stklcode: Re-added "edit_user_profile_update" for now in decdd88a4d273f4fb5b7e73f8468ba8d26750ede

We can discuss about refining this in future changes.

inc/class-cachify.php
stklcode: resolved in feab290d58cb32083c8fd2ae105930cab2dac7d3

inc/class-cachify.php
stklcode: Yes, we could. In the meantime we raised the requirements to 4.7 anyway, so I guess that's no longer a point here.

stklcode (approved):
Should be OK now.

A second opinion would not hurt here though.

PR #225, opened 2021-04-18
Flush on saved or updated posts overwork

timse201:
+ added some third party
#223


inc/class-cachify.php
```diff
@@ -863,70 +854,65 @@ public static function touch_comment( $new_status, $old_status, $comment ) {
 	}
 
 	/**
-	 * Generate publish hook for custom post types
+	 * Removes the post type cache if saved or updated
 	 *
-	 * @since   2.1.7  Make the function public
-	 * @since   2.0.3
+	 * @since
+	 * @change
```
stklcode: Was the change history removed by intention?

Despite `@change` not being a valid PHPdoc tag (even though this project makes excessive use of it), empty `@since` does not make any sense either.

```php
@since 2.0.3 [implicit: created]
@since 2.1.7 Make the function public
```

perfectly points out relevant changes to this method...

inc/class-cachify.php
```diff
@@ -863,70 +854,65 @@ public static function touch_comment( $new_status, $old_status, $comment ) {
...
-			array(
-				'public' => true,
-			)
-		);
 
-		/* Empty data? */
-		if ( empty( $post_types ) ) {
-			return;
-		}
+	public static function save_update_trash_post( $id ) {
```
stklcode: Renaming a public function without prior deprecrecation and not even a change note (e.g. third `@since` tag) is not really kind.

So we should either keep and deprecate the old method and introduce a new one or raise the project version to 3.0 (prior deprecation would be nice though)

inc/class-cachify.php
```diff
@@ -863,70 +854,65 @@ public static function touch_comment( $new_status, $old_status, $comment ) {
...
 	}
 
 	/**
-	 * Removes the post type cache on post updates
+	 * Removes the post type cache before an existing post type is updated in the db
 	 *
-	 * @since   2.0.3
-	 * @change  2.3.0
+	 * @since
+	 * @change
```
stklcode: see above

inc/class-cachify.php
```diff
@@ -863,70 +854,65 @@ public static function touch_comment( $new_status, $old_status, $comment ) {
...
-	public static function publish_post_types( $post_id, $post ) {
-		/* No post_id? */
-		if ( empty( $post_id ) || empty( $post ) ) {
-			return;
-		}
 
-		/* Post status check */
-		if ( ! in_array( $post->post_status, array( 'publish', 'future' ), true ) ) {
-			return;
+	public static function post_update( $id, $data ) {
```
stklcode, line 1037: see above

stklcode (changes requested):

inc/class-cachify.php
```diff
@@ -863,70 +854,65 @@ public static function touch_comment( $new_status, $old_status, $comment ) {
...
+	 *
+	 * @since
+	 * @change
+	 *
+	 * @param	integer|object	$data  Post ID or post
+	 */
+
+	public static function flush_cache_for_posts( $data ) {
+
+		if ( is_int( $post ) ) {
```
stklcode, line 1055: Guess the variable `$data`  should be used here or the parameter must be renamed to `$post`.

stklcode: fixed in ec1c67a591fbe5d25975ef04de0e6c35238c564a

inc/class-cachify.php
stklcode: resolved in 7bd42ba7e492bae695062a53f632f91bc0fe2ab1

inc/class-cachify.php
stklcode: resolved in 7bd42ba7e492bae695062a53f632f91bc0fe2ab1

stklcode (approved):

stklcode:
LGTM

Gave it a quick test and it seems flushing works as expected without breaking anything. Should be part of pre-release testing routine again.

PR #226, opened 2021-04-20
change composer action for deployment workflows

### LINKED ISSUE #218: 2 unnecessary files
docker_tag and output.log are pushed to wordpress.org

stklcode:
The previous composer action adds some extra files to the project root. This PR resolves #218.

Those can be ignored, but the action used for testing is fast and allows to specify the exact PHP version whereas the composer action uses the latest stable (we stumbled upon this in other projects).

The `shivammathur/setup-php@v2` version has proven to work since #217, so why not use the same workflow in all 3 recipes...

### DIFF
--- a/.github/workflows/wordpress-plugin-asset-update.yml
+++ b/.github/workflows/wordpress-plugin-asset-update.yml
@@ -9,7 +9,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: php-actions/composer@v5
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.0'
+          tools: composer
+      - name: Build
+        run: composer install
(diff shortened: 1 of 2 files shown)


Zodiac1978:
Can anyone with composer knowledge/opinion help here? @pfefferle, @2ndkauboy , @florianbrinkmann  or @patrickrobrecht ? If a second pair of eyes had a look I can merge this.

pfefferle:
What about downloading the required composer as phar version?

pfefferle:
And/or it seems possible to downgrade composer with for example `composer self-update 1.4.1` to any version.

stklcode:
The action in use is capable of doing that, if necessary ([docs](https://github.com/shivammathur/setup-php#wrench-tools-support))

Just use `tools: composer:v2`, `tools: composer:2.1.3` or similar instead of `tools: composer`.

patrickrobrecht (commented):
LGTM.

PR #229, opened 2021-05-12
use default styling for "flush cache now" button (#227)

### LINKED ISSUE #227: "Flush cache now" button contrast
The "Flush cache now" button has a orange background with white text. This is a small contrast ratio of 2.86.

see: https://colourcontrast.cc/dd823b/ffffff

stklcode:
resolves #227 

The orange button has suboptimal color contrast and indicates some kind of dangerous behavior, which is not the case here. Use default style class for secondary buttons instead.

This obsoletes the _settings.css_ file.

I have refrained from adding the trash-icon to the button like proposed in the issue, because this is rarely done across WP settings pages and positioning the icon correctly inside the button required additional CSS again. There's a trash can one line above and in the admin bar, 3 seems somewhat overdone.


krafit (approved):
LGTM, thank you!

PR #231, opened 2021-05-19
chore: add cachify_buffer filter

matzeeable:
Hi!

Here is the developer of [Real Cookie Banner](https://wordpress.org/plugins/real-cookie-banner/) and I want to make our plugin compatible with your caching solution. Unfortunately, your cache does save the output buffer "too early", so we cannot hook into.

We intend to "block" content until a user gives consent, so we need to alter the HTML code before it gets cached.

I would like to propose the following filter: `cachify_buffer`.

What do you think?

Regards,
Matthew 😊

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1337,14 +1337,14 @@ final class Cachify {
 			/**
 			 * Filters the buffered data itself
 			 *
-			 * @since 2.3.2
+			 * @since 2.4
 			 *
 			 * @param string $data          The actual data.
 			 * @param object $method        Instance of the selected caching method.
 			 * @param string $cache_hash    The cache hash.
 			 * @param int    $cache_expires Cache validity period.
 			 */
-			$data = apply_filters( 'cachify_buffer', $data, self::$method, self::_cache_hash(), self::_cache_expires() );
+			$data = apply_filters( 'cachify_modify_output', $data, self::$method, self::_cache_hash(), self::_cache_expires() );
 
 			call_user_func(
 				array(
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1334,6 +1334,18 @@ final class Cachify {
 
 		/* Save? */
 		if ( $should_cache ) {
+			/**
+			 * Filters the buffered data itself
+			 *
+			 * @since 2.3.2
+			 *
+			 * @param string $data          The actual data.
+			 * @param object $method        Instance of the selected caching method.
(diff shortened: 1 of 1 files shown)


inc/class-cachify.php
```diff
@@ -1334,6 +1334,18 @@ public static function set_cache( $data ) {
 
 		/* Save? */
 		if ( $should_cache ) {
+			/**
+			 * Filters the buffered data itself
+			 *
+			 * @since 2.3.2
```
stklcode: Definitely out of scope for a patch release, should be 2.4 or whatever point release this will be merged into.

inc/class-cachify.php
```diff
@@ -1334,6 +1334,18 @@ public static function set_cache( $data ) {
...
+			 *
+			 * @param string $data          The actual data.
+			 * @param object $method        Instance of the selected caching method.
+			 * @param string $cache_hash    The cache hash.
+			 * @param int    $cache_expires Cache validity period.
+			 */
+			$data = apply_filters( 'cachify_buffer', $data, self::$method, self::_cache_hash(), self::_cache_expires() );
```
stklcode: The name `cachify_buffer` is not quite self-explaining (there is no such thing like a "buffer" from an outside perspective). I'd prefer something like `cachify_pre_store` or `cachify_modify_output`.

stklcode (commented):
Hi Matthew, thanks for your PR!

I don't see anything against the proposed hook. Left some minor remarks on the naming, but nothing substantial.

We should check if the changes from #93 that have been partially reverted before the 2.3 release (#170) would mess up with the intended use case, if they might be reintroduced in a future release, but for now it seems OK.

inc/class-cachify.php
matzeeable: Good point! Changed it to `cachify_modify_output`.

matzeeable:
Hey @stklcode !

Thanks for your review. Just commited the changes, please give it a try 😊

matzeeable:
Hey @stklcode !

Did you have a chance to have a look at this? 😊

florianbrinkmann (approved):
LGTM, thanks for the PR @matzeeable

matzeeable:
Hey @florianbrinkmann !

I just wanted to ask when the changes will be published?

florianbrinkmann:
Hi @matzeeable, there are a few tickets left in the milestone for 2.4.0, that need to be closed https://github.com/pluginkollektiv/cachify/milestone/4

PR #241, opened 2021-08-15
apc: rework sanitization to not use WP functions (#240)

### LINKED ISSUE #240: Cachify is broken for APC with Apache with cachify 2.3.1
When [using APC with Apache as described in the documentation, one would add these lines](https://cachify.pluginkollektiv.org/documentation/#htaccess-example-apache) to the `.htaccess` file for Wordpress:

```
<Files index.php>
    php_value auto_prepend_file /absolute path to/plugins/cachify/apc/proxy.php
</Files>
```

Within the referenced file `proxy.php`, there are calls to the [wordpress core function `wp_unslash`](https://developer.wordpress.org/reference/functions/wp_unslash/). Unfortunately, at the time the file `proxy.php` is called and executed by the PHP interpreter, the Wordpress files have not been executed yet and the Wordpress core functions are not available. 

This results in a fatal PHP error and and HTTP 500 error to the client; the page is not accessible anymore. In the log file, the following error message can be found:

```
[Sun Aug 15 15:27:13.675187 2021] [php7:error] [pid 7300] [client <some-ip>:49347] PHP Fatal
error:  Uncaught Error: Call to undefined function wp_unslash() in /var/www/<domain>/wordpress/wp-content/plugins/cachify/apc/proxy.php:27\nStack trace:\n#0 /var/www/<domain>/wordpress/wp-content/plugins/cachify/apc/proxy.php(47): cachify_is_ssl()\n#1 {main}\n  thrown in /var/www/<domain>/wordpress/wp-content/plugins/cachify/apc/proxy.php on line 27, referer: https://www.<domain>/
```

As you can see in the git history, these changes have been introduced on February7th 2021:

https://github.com/pluginkollektiv/cachify/commit/f8326db0d05467c9041ca9f80b619274f57efb12#diff-42464b97adc53329b9e4cf07eaf86c73735732e2cae87c3f6cc9a9ba629f4831

**Proposed solution: Revert calls to Wordpress core functions in APC proxy file.**

Also please note, depending on the configuration of cachify, this is only visible for anonymous user since calls to cachify are skipped for logged-in users.

stklcode:
fixes #240

WordPress is not initialized when the APC proxy is called, so we must not rely on logic like `wp_unslash().`

Use `filter_input` to sanitize untrusted data and drop the unslashing, as it is not necessary in this place.

### DIFF
--- a/apc/proxy.php
+++ b/apc/proxy.php
@@ -23,12 +23,8 @@ if ( ! empty( $_COOKIE ) ) {
  */
 function cachify_is_ssl() {
 	if ( isset( $_SERVER['HTTPS'] ) ) {
-		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-		if ( 'on' === strtolower( wp_unslash( $_SERVER['HTTPS'] ) ) ) {
-			return true;
-		}
-
-		if ( '1' === $_SERVER['HTTPS'] ) {
+		$https = filter_input( INPUT_SERVER, 'HTTPS', FILTER_SANITIZE_STRING );
+		if ( 'on' === strtolower( $https ) || '1' === $https ) {
 			return true;
 		}
 	} elseif ( isset( $_SERVER['SERVER_PORT'] ) && ( '443' === $_SERVER['SERVER_PORT'] ) ) {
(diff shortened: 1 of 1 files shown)


pfefferle (approved):
LGTM

PR #246, opened 2021-11-16
Update deprecated WPMS hooks

ouun:
`delete_blog` & `wpmu_new_blog` are deprecated since WP 5.1. This PR follows up.

References:
https://developer.wordpress.org/reference/hooks/wpmu_new_blog/
https://developer.wordpress.org/reference/hooks/delete_blog/


inc/class-cachify.php
```diff
@@ -371,21 +371,21 @@ public static function on_uninstall() {
...
+	 * Uninstalling of the plugin for WPMS site.
 	 *
 	 * @since   1.0
 	 * @change  1.0
 	 *
-	 * @param integer $id  Blog ID.
+	 * @param WP_Site $old_site Old Site object.
 	 */
-	public static function uninstall_later( $id ) {
+	public static function uninstall_later( $old_site ) {
```
stklcode, line 296: The change for `install_later` add the type hint ` WP_Site $new_site`. Was this omitted in this place intentionally?

However, this is strictly speaking public API. So I'd suggest to make it backwards compatible for now, i.e. drop the type hint and use implicit union type `int|WP_Site`. We only need the ID anyway, so we can simply extract it, if we get a `WP_Site`  instance and passthrough integers.

E.g. something like 
```php
switch_to_blog( is_int( $new_site ) ? $new_site : $new_site->blog_id );
```

stklcode (changes requested):
Hi @ouun, thanks for your PR.

The changes are absolutely reasonable. I left a minor remark in the argument types though.

stklcode:
Another decision that should at least be discussed briefly:

[`wp_initialize_site`](https://developer.wordpress.org/reference/functions/wp_initialize_site/) and [`wp_delete_site`](https://developer.wordpress.org/reference/functions/wp_delete_site/) have both been introduced in 5.1 - _Cachify_ currently declares compatibility with 4.4+

So we either drop support for WP <5.1 or add both hooks, maybe conditionally.

florianbrinkmann:
> So we either drop support for WP <5.1 or add both hooks, maybe conditionally.

I would vote for adding both and checking for the WordPress version.

inc/class-cachify.php
```diff
@@ -134,21 +134,14 @@ public function __construct() {
...
-					__CLASS__,
-					'uninstall_later',
-				)
-			);
+			if ( version_compare( get_bloginfo( 'version' ), '5.1', '<' ) ) {
+				// The following hooks are deprecated since WP 5.1 (#246).
+				add_action( 'wpmu_new_blog', array( __CLASS__, 'install_later' ) );
+				add_action( 'delete_blog', array( __CLASS__, 'uninstall_later' ) );
+			} else {
+				add_action( 'wp_initialize_site', array( __CLASS__, 'install_later' ) );
```
stklcode, line 140: Do we agree that [`wp_initialize_site`](https://developer.wordpress.org/reference/hooks/wp_initialize_site/) is the correct hook here?

We used to use [`wpmu_new_blog`](https://developer.wordpress.org/reference/hooks/wpmu_new_blog/) which "fires immediately after a new site is created".
The recommended replacement is [`wp_insert_site`](https://developer.wordpress.org/reference/hooks/wp_insert_site/) which "fires once a site has been inserted into the database".

We do add empty options (i.e. all default) into the database. Can be seen as part of the "initialization routine", so the initialize-hook might be the semantically better choice here. Just wanna make sure this is done intentionally.

inc/class-cachify.php
```diff
@@ -261,21 +266,21 @@ public static function on_uninstall() {
...
-	public static function uninstall_later( $id ) {
+	public static function uninstall_later( $old_site ) {
 		/* No network plugin */
 		if ( ! is_plugin_active_for_network( CACHIFY_BASE ) ) {
 			return;
 		}
 
 		/* Switch to blog */
-		switch_to_blog( $id );
+		switch_to_blog( is_int( $new_site ) ? $new_site : $new_site->blog_id );
```
stklcode: Guess the argument should be `$old_site` here.

stklcode (approved):
Just fixed a small copy&paste bug and added some cosmetic changes to the PHPdocs.

I'm fine with the changes.

PR #247, opened 2021-11-16
Separate Hooks: Flushing Full Site Cache vs. Removing Single Post Cache

### LINKED ISSUE #245: wp_delete_post_revision() flushes complete cache
**Describe the bug**
Thanks for the ongoing work on this simple and great plugin.
I just realized that the automatic processing function [`wp_delete_post_revision()`](https://developer.wordpress.org/reference/functions/wp_delete_post_revision/) calls [`wp_delete_post()`](https://developer.wordpress.org/reference/functions/wp_delete_post/) to delete the revisions. This **flushes the complete Cachify cache** as `do_action('before_delete_post')` [is triggered](https://github.com/pluginkollektiv/cachify/blob/b47f58a39daa4cece961706fcf81c2e2b5c4c2ea/inc/class-cachify.php#L1106).

This flushes the complete page cache unnecessarily as post revisions won't get cached at all. Deleting them has no impact on the frontend.

**To Reproduce**
1. Limit revisions in wp-config: `define('WP_POST_REVISIONS', 1)`
1. Edit a post that has already revisions
2. Note that `wp_delete_post_revision()` is executed
3. Check your cache
4. Realize it is gone

**Expected behavior**
Cache flushes only for posts with status "publish". The hook `before_delete_post` can pass `$post_id` and `$post` to the `flush_total_cache()` function. A check for the post status could help, here. Please let me know if you are open for a PR into this direction.

**System (please complete the following information):**
- Version 2.3.2

ouun:
Related to: https://github.com/pluginkollektiv/cachify/issues/245

This PR is a draft to avoid unnecessary caching cycles. Cachify currently flushes the full page cache for the single post related hooks `before_delete_post` & `wp_trash_post`. My intention is to separate such hooks into a more specific cache flushing route that is limited to the actual post.

I see the possible downside of that approach as a deleted post could be listed in an archive, too. So deleting only the post cache might lead to 404s. So this needs your feedback and suggestions.

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -92,8 +92,11 @@ final class Cachify {
 			99
 		);
 
-		/* Flush Hooks */
+		/* Flush Full Cache Hooks */
 		add_action( 'init', array( __CLASS__, 'register_flush_cache_hooks' ), 10, 0 );
+
+        /* Remove Post Cache Hooks */
+        add_action( 'init', array( __CLASS__, 'register_remove_post_cache_hooks' ), 10, 0 );
 
 		add_action(
 			'cachify_remove_post_cache',
@@ -941,13 +944,13 @@ final class Cachify {
 	 * Removes a page (id) from cache
 	 *
 	 * @since   2.0.3
-	 * @change  2.1.3
+	 * @change  2.4.0
 	 *
 	 * @param   integer $post_id  Post ID.
 	 */
 	public static function remove_page_cache_by_post_id( $post_id ) {
 		$post_id = (int) $post_id;
-		if ( ! $post_id ) {
+		if ( ! $post_id || ! in_array( get_post_status( $post_id ), array( 'publish', 'future' ), true ) ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


florianbrinkmann:
Hi @ouun,

thanks for opening the PR! I think it’s a great idea to prevent that cache flush from happening when a revision is deleted. But only flushing the post cache on post removal is – in my opinion – not a way we should go, and you already mentioned that could lead to problems.

I searched a bit if we could check in `flush_total_cache()` if the current request is for a post revision, and came up with the following lines, added at the beginning of the `flush_total_cache()` method:

```php
$saved_post_revision = did_action( 'save_post_revision' );
if ( $saved_post_revision ) {
	// We do not need to flush the cache for updated post revisions.
	return;
}
```

That worked in my tests, the complete cache was only flushed once for a post update. I created a branch with the change (https://github.com/pluginkollektiv/cachify/tree/do-not-flush-cache-for-post-revisions), would you give it a try and see if that fixes the issue?

Thanks!

florianbrinkmann:
#261 handles that without possible edge cases

PR #251, opened 2021-12-23
Fix issue with port in URL

### LINKED ISSUE #250: URL with port number leads to issue with cache removal
**Describe the bug**
I am using `localhost` with different port numbers for local development and just noticed that removing a cache by URL does not work. The reason for this is that the port is part of the string that is hashed when a new cache transient is created. But when the transient is searched to remove it, it looks for the hash generated from a string without the port.

**To Reproduce**
1. Use a domain with a port number.
2. Fill the cache for a post.
3. Add a comment to the post and approve it.
4. Now, when not being logged in, the comment should not be visible because the cached version is loaded.

florianbrinkmann:
Fixes #250.

The problem with the current code is, that the generated hash from `Cachify::_cache_hash()` is different for URLs with a port, depending on if the `$url` method param is empty or not. The reason for that is, that the port is part of `$_SERVER['HTTP_HOST']` but not `$url_parts['host']`.

This PR modifies the code, so that, if `$url` is empty, the URL from the `$_SERVER` values is saved in `$url`, and the generation of the string for the hash is equal for both cases.

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1006,6 +1006,7 @@ final class Cachify {
 	 *
 	 * @since   0.1
 	 * @change  2.0
+	 * @change  2.4.0 Fix issue with port in URL.
 	 *
 	 * @param   string $url  URL to hash [optional].
 	 * @return  string       Cachify hash value.
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1012,15 +1012,14 @@ final class Cachify {
 	 */
 	private static function _cache_hash( $url = '' ) {
 		$prefix = is_ssl() ? 'https-' : '';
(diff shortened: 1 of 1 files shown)


stklcode (approved):
The change does harmonize the behavior, i.e. eliminates the port in the cache key, if present.

One could imagine some scenario where the port makes a difference and two sites, i.e. "https://example.com/" and "https://example.com:8443/" have different contents and share one caching backend.

I think this is as unlikely as an "https" prefix collision where "https://example.com/" and "http://https-example.com/" would share the same caching key "https-example.com", so I'm fine with this change.

florianbrinkmann:
Thanks for your review @stklcode!

PR #254, opened 2022-01-07
Exclude sitemap.xml from caching

### LINKED ISSUE #242: WordPress Core sitemaps are cached (and served with the wrong content type when using HDD caching)
**Describe the bug**
Since Wordpress 5.5 there is [a sitemap feature included in Core](https://make.wordpress.org/core/2020/07/22/new-xml-sitemaps-functionality-in-wordpress-5-5/). When using HDD Caching those sitemaps will be served as text/html (as they are cached in a .html file) instead of the correct type application/xml.

**To Reproduce**
1. Install Cachify
2. Enable HDD caching & add the provided snippet to .htaccess
3. Visit `/wp-sitemap.xml` (Content-type: application/xml)
4. Reload page, it should display the cached version with the wrong content type (=> Content-type: text/html)

**Expected behavior**
Sitemaps are either not cached at all or served with the correct content type.

Currently RSS feeds are not cached, so sitemaps should probably be excluded to exclude all core-generated non-html content?

**System (please complete the following information):**
- Version 2.3.1 / WordPress 5.8.1

**Additional context**
The following code works to detect all sitemaps and the corresponding XSL style sheets:

```
add_filter(
	'cachify_skip_cache',
	function() {
		global $wp_query;
		return ! empty( $wp_query->query_vars["sitemap"] ) ||
			! empty( $wp_query->query_vars["sitemap-subtype"] ) ||
			! empty( $wp_query->query_vars["sitemap-stylesheet"] );
	}
);
```

florianbrinkmann:
Fixes #242.

The PR checks for the new sitemap feature added in WP 5.5, thanks @ManuelHu for the query var checks in the issue and @chesio for the tip with `get_query_var()`.

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1135,6 +1135,7 @@ final class Cachify {
 	 *
 	 * @since   0.2
 	 * @change  2.3.0
+	 * @change  2.4.0 Add check for sitemap feature.
 	 *
 	 * @return  boolean              TRUE on exclusion
 	 *
@@ -1200,6 +1201,11 @@ final class Cachify {
 					return true;
 				}
 			}
+		}
+
+		// Sitemap feature added in WP 5.5.
+		if ( get_query_var( 'sitemap' ) || get_query_var( 'sitemap-subtype' ) || get_query_var( 'sitemap-stylesheet' ) ) {
+			return true;
 		}
 
 		return false;


stklcode:
Serving the correct content types would be a more elegant solution to the actual problem. There’s no reason not to cache sitemaps or feeds in general.

This might be achieved via webserver config for static HDD cache, derived from the request URI. Well, possible,  but not elegant.

But since _Cachify_ has no logic for non-HTML content and the clearing mechanisms would have to be adjusted to properly invalidate the sitemap cache, exclusion seems reasonable at this point.

stklcode (approved):

krafit (approved):

PR #255, opened 2022-01-26
Add HDD Cache expiration cron

### LINKED ISSUE #61: add a cron - hdd cache
due to technical limitations the hdd cache can not expire
these days some small hoster have a webinterface to add crons, too

so adding a cron inside the plugin would be helpful
if the user wants to use this option there should be a small notification that wp-cron.php should get hitted by a server cron or a service provider

when selecting 0 the cron should be disabled

~~until that the option should be disabled and there should be an information while switching to hdd~~ (#62)

#32

Ancocodet:
This PR adds an wp_cron for clearing the hdd cache.

The option for changing the expiration time is reenabled when hdd is selected.
On the Settings page a small information is added which informs the user that an system cron should be used (only when hdd is selected).

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -144,7 +144,7 @@ final class Cachify {
 
 			$timestamp = wp_next_scheduled( 'hdd_cache_cron' );
 			if ( $timestamp == false ) {
-				wp_schedule_event( time(), 'cache_expire', 'hdd_cache_cron' );
+				wp_schedule_event( time(), 'cashify_cache_expire', 'hdd_cache_cron' );
 			}
 
 			add_action(
@@ -563,7 +563,7 @@ final class Cachify {
 	 * @return array
 	 */
 	public static function add_cron_cache_expiration() {
-		$schedules['cache_expire'] = array(
+		$schedules['cashify_cache_expire'] = array(
 			'interval' => self::$options['cache_expires'] * 3600,
 			'display'  => 'Cache Expiration',
 		);
@@ -830,7 +830,7 @@ final class Cachify {
 		if ( self::METHOD_HDD == self::$options['use_apc'] ) {
 			$timestamp = wp_next_scheduled( 'hdd_cache_cron' );
 			if ( $timestamp !== false ) {
-				wp_reschedule_event( $timestamp, 'cache_expire', 'hdd_cache_cron' );
+				wp_reschedule_event( $timestamp, 'cashify_cache_expire', 'hdd_cache_cron' );
 				wp_unschedule_event( $timestamp, 'hdd_cache_cron' );
 			}
 		}
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -562,7 +562,7 @@ final class Cachify {
 	 *
 	 * @return array
 	 */
-	public static function add_cron_cache_expiration() {
+	public static function add_cron_cache_expiration( $schedules ) {
 		$schedules['cachify_cache_expire'] = array(
 			'interval' => self::$options['cache_expires'] * 3600,
 			'display'  => esc_html__( 'Cachify expire', 'cachify' ),
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -565,7 +565,7 @@ final class Cachify {
 	public static function add_cron_cache_expiration() {
 		$schedules['cashify_cache_expire'] = array(
 			'interval' => self::$options['cache_expires'] * 3600,
-			'display'  => __( 'Cachify expire', 'cachify' ),
+			'display'  => esc_html__( 'Cachify expire', 'cachify' ),
 		);
 		return $schedules;
 	}
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -565,7 +565,7 @@ final class Cachify {
 	public static function add_cron_cache_expiration() {
 		$schedules['cashify_cache_expire'] = array(
 			'interval' => self::$options['cache_expires'] * 3600,
-			'display'  => 'Cache Expiration',
+			'display'  => 'Cachify',
 		);
 		return $schedules;
 	}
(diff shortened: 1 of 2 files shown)


inc/class-cachify.php
```diff
@@ -132,6 +132,30 @@ public function __construct() {
 			2
 		);
 
+		/* Add Cron for clearing the HDD Cache */
+		if ( self::METHOD_HDD == self::$options['use_apc'] ) {
```
bueltge: We would like to use more strict code, === instead of ==
please check and validate them, enhance the PR in this topic. Thx.

inc/class-cachify.php
```diff
@@ -132,6 +132,30 @@ public function __construct() {
...
+			add_filter(
+				'cron_schedules',
+				array(
+					__CLASS__,
+					'add_cron_cache_expiration',
+				)
+			);
+
+			$timestamp = wp_next_scheduled( 'hdd_cache_cron' );
+			if (false == $timestamp) {
```
bueltge: Same here, === instead of ==

inc/class-cachify.php
```diff
@@ -270,6 +294,14 @@ public function __construct() {
 	 * @change  2.1.0
 	 */
 	public static function on_deactivation() {
+		/* Remove hdd cache cron when hdd is selected */
+		if ( self::METHOD_HDD == self::$options['use_apc'] ) {
```
bueltge: Same, === instead of ==

inc/class-cachify.php
```diff
@@ -514,6 +546,30 @@ public static function robots_txt( $data ) {
...
+	 * Add cache expiration cron shedule
+	 *
+	 * @since  2.3.3
+	 *
+	 * @return array
+	 */
+	public static function add_cron_cache_expiration() {
+		$schedules['cashify_cache_expire'] = array(
+			'interval' => self::$options['cache_expires'] * 3600,
+			'display'  => 'Cachify',
```
bueltge: The string should be translatable.

inc/class-cachify.php
```diff
@@ -769,6 +825,16 @@ public static function process_flush_request( $data ) {
 				);
 			}
 		}
+
+		/* Reschedule HDD Cache Cron */
+		if ( self::METHOD_HDD == self::$options['use_apc'] ) {
```
bueltge: same, === instead ==

inc/class-cachify.php
```diff
@@ -769,6 +825,16 @@ public static function process_flush_request( $data ) {
...
+			$timestamp = wp_next_scheduled( 'hdd_cache_cron' );
+			if ( false != $timestamp ) {
```
bueltge: same, !== instead of !=

bueltge (changes requested):
Thanks a lot for the PR; great work. Please look at my notes so that we get an update from you to solve the small topics.

inc/class-cachify.php
Ancocodet: Is included in "Make requested changes to checks"

inc/class-cachify.php
Ancocodet: Is included in "Make requested changes to checks"

inc/class-cachify.php
Ancocodet: Is included in "Make requested changes to checks"

inc/class-cachify.php
Ancocodet: Is included in "Make requested changes to checks"

inc/class-cachify.php
Ancocodet: Is included in "Make requested changes to checks" and found the same issue in line 300

inc/class-cachify.php
Ancocodet: I'm not sure if this is really needed, because this is the shedule name, but i changed it in my last commit.

inc/class-cachify.php
```diff
@@ -514,6 +546,30 @@ public static function robots_txt( $data ) {
...
+	 * Add cache expiration cron shedule
+	 *
+	 * @since  2.3.3
+	 *
+	 * @return array
+	 */
+	public static function add_cron_cache_expiration() {
+		$schedules['cashify_cache_expire'] = array(
+			'interval' => self::$options['cache_expires'] * 3600,
+			'display'  => __( 'Cachify expire', 'cachify' ),
```
bueltge: Small note, needs sanitization like `esc_html__()`

bueltge (changes requested):

inc/class-cachify.php
Ancocodet: Done

bueltge (approved):

pfefferle (approved):
LGTM

inc/class-cachify.php
```diff
@@ -132,6 +132,30 @@ public function __construct() {
...
+				'cron_schedules',
+				array(
+					__CLASS__,
+					'add_cron_cache_expiration',
+				)
+			);
+
+			$timestamp = wp_next_scheduled( 'hdd_cache_cron' );
+			if ( false === $timestamp ) {
+				wp_schedule_event( time(), 'cashify_cache_expire', 'hdd_cache_cron' );
```
stklcode: What about ... let's say `cachify` instead of `cashify`? ;)

inc/class-cachify.php
```diff
@@ -514,6 +546,30 @@ public static function robots_txt( $data ) {
...
+	}
+
+	/**
+	 * Add cache expiration cron shedule
+	 *
+	 * @since  2.3.3
+	 *
+	 * @return array
+	 */
+	public static function add_cron_cache_expiration() {
```
stklcode: The method signature is missing the  `$schedules` parameter, so it always creates a new single-element array.

stklcode (changes requested):

inc/class-cachify.php
Ancocodet: Missed that, fixed it now

stklcode (approved):

Issue #99, opened 2017-05-24
Clean up the use of @ operator


chesio:
Inspired by discussion in #98. Here's a list of current uses of `@` operator in the plugin:
- [x] [inc/cachify.class.php:1596-1598](https://github.com/pluginkollektiv/cachify/blob/master/inc/cachify.class.php#L1596-1598)
- [ ] [inc/class-cachify-apc.php:103](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-apc.php#L103)
- [x] [inc/class-cachify-hdd.php:115](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-hdd.php#L115)
- [ ] [inc/class-cachify-hdd.php:188](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-hdd.php#L188)
- [x] [inc/class-cachify-hdd.php:195](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-hdd.php#L195)
- [ ] [inc/class-cachify-hdd.php:200](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-hdd.php#L200)
- [x] [inc/class-cachify-hdd.php:203](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-hdd.php#L203)
- [ ] [inc/class-cachify-hdd.php:254](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-hdd.php#L254)
- [x] [inc/class-cachify-memcached.php:140](https://github.com/pluginkollektiv/cachify/blob/2abed98e1029729ab55a1d5e8b5723176213070c/inc/class-cachify-memcached.php#L140)

Let's try to reduce this list to necessary minimum :-)

angcl:
Perhaps this issue could be closed now because #256 was merged.

stklcode:
The _master_ references in the first comment pointed to inexisting files. I've just updated them by comparing the code from 05/2017 to the current _develop_ state with fixed commit references.

[...]

PR #256, opened 2022-02-01
Remove unnecessary @ operators

florianbrinkmann:
This is a follow-up to #99. @angelocali and I looked at the remaining `@` in the code and identified two that can be removed. The other functions prefixed with the operator throw warnings in case of failure regarding to the docs, whatever that means.


pfefferle (approved):
LGTM

bueltge (approved):

stklcode (approved):

PR #260, opened 2022-02-03
use one-line notation for add_action calls

stklcode:
The initialization routines got pretty long and unhandy to read with a mixture of `add_*` blocks and conditionals.

Re-format the code to use one-line notations which do not exceed 96 characters (effectively 102 with tab-size 4) each and compress the class by a total of ~164~ 176 lines.

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -83,208 +83,66 @@ final class Cachify {
 		self::$is_nginx = $GLOBALS['is_nginx'];
 
 		/* Publish hooks */
-		add_action(
-			'init',
-			array(
-				__CLASS__,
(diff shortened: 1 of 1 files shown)


pfefferle:
But won't the `add_action` one-liners be marked as issues by PHPCS and/or auto-fixed to multi-line by PHPCBF?

stklcode:
At least not with our current ruleset (which is simply `<rule ref="WordPress"/>`) - obviously, "quality" CI is passing.

With warnings enabled there are 27 remaining issues in _class-cachify.php_ after auto-fixing, but `add_action` indentation is none of them.

pfefferle (approved):
LGTM

PR #262, opened 2022-02-08
Add filter to disable gzip file creation

### LINKED ISSUE #197: add option to disable GZ compression for HDD caching
Related to request from WP support forums: https://wordpress.org/support/topic/disable-gz/

The HDD caching module always generates a static HTML file along with a gzipped version:
https://github.com/pluginkollektiv/cachify/blob/dcac6b7babdef4454782faf6f9c076699cb11944/inc/cachify_hdd.class.php#L165-L166

We have a corresponding block in the Apache HTTP server example using _mod_mime_ to server the pre-compressed content.

With most webservers supporting compression natively which can be superior to compression by PHP and more important is not handled during initial site visit, many sites don't need the explicit _*.gz_ files.

We should at least introduce an option to disable generation of pre-compressed files to reduce the overhead here.

angcl:
@florianbrinkmann and I worked on this PR that adds a filter that allows to disable gzip file creation.

Fixes #197

### DIFF
--- a/inc/class-cachify-hdd.php
+++ b/inc/class-cachify-hdd.php
@@ -170,14 +170,7 @@ final class Cachify_HDD {
 		}
 		/* Write to file */
 		self::_create_file( self::_file_html( $file_path ), $data );
-
-		/**
-		 * Filter to disable creation of GZIP files.
-		 *
-		 * @param bool $disable_gzip
-		 */
-		$disable_gzip = apply_filters( 'cachify_disable_gzip_file_creation', false );
-		if ( ! $disable_gzip ) {
+		if ( apply_filters( 'cachify_create_gzip_files', true ) ) {
 			self::_create_file( self::_file_gzip( $file_path ), gzencode( $data, 9 ) );
 		}
 	}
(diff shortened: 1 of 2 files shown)


inc/class-cachify-hdd.php
```diff
@@ -170,6 +170,12 @@ private static function _create_files( $data ) {
 		}
 		/* Write to file */
 		self::_create_file( self::_file_html( $file_path ), $data );
+
+		/**
+		 * Filter that allows to enable/disable GZIP file creation
+		 *
+		 * @param bool $create_gzip_files Whether to create GZIP files. Default is `true`
```
pfefferle: shouldn't it be `@return`?

pfefferle (approved):
LGTM!

stklcode:
Filter looks good.

We should conditionally remove the GZip instructions from the .htaccess instructions in the settings tab, too, if generation is disabled, even though this should be obvious … but we all know copy&pasting is one thing, understanding another.

Did we discusse a checkbox in the settings that may address more users?

florianbrinkmann:
I guess that is a feature that is only relevant for a small part of our users, so that a filter should be enough, but that is only a guess :)

angcl:
gzip instructions now won't be added to .htaccess instructions if gzip file creation is disabled. :)

stklcode (approved):

Issue #259, opened 2022-02-03
Raise required WordPress and PHP version


pfefferle:
WordPress needs at least PHP version 5.6.20+ (but recommends 7.4), so I would recommend to raise the required PHP version of cachify to 5.6 (and the required WordPress version accordingly).

-- https://wordpress.org/about/requirements/

stklcode:
+1 for not sticking to legacy PHP.

Your suggestion for a reasonable minimum WP version? We currently support down to 4.4.
REST API stuff (like #257) requires 4.7 anyway.

PHP 5.6 is required since WP 5.2. Older releases are backwards compatible to PHP 5.2
We can raise the requirements independently, personally I'd be fine with 5.6/4.7 (but don't have requirements for anything below WP 5.8)

florianbrinkmann:
Oh, I thought the REST API was in core since 4.4 because `register_rest_route` was introduced there, but you are right, it was 4.7.

How would we raise the versions? Add a note to a release that we will raise it in the next release, so people with older versions are informed about it, or just raise it from one version to the other?

krafit:
My 2 cents: I'd raise to 5.6/4.7 in 2.4.0.
It's safe to assume, users with super old PHP / WP versions don't update their plugins regularly.

stklcode:
I didn’t check the code, maybe our route is already working with 4.4. We only need a very small fraction of the API features. Officially it’s 4.7.

If you are running such version not just because yo don’t care about the site, there are usually reasons for that like major customizations that cannot be simply adjusted and designed to live until the core is EOL (3.7 received security updates, so it could still be used for site core). Plugin updates, if any, usually require manual inspection anyway in that case.

florianbrinkmann:
@angelocali and I just tried #257 with 4.4 and it worked (except a missing dashicon)

PR #267, opened 2022-05-17
changed $should_cache default

### LINKED ISSUE #266: Cachify caches output from WordPress with HTTP-Status-Code 500, but shouldnt
We are using the nginx-hdd Version of cachify and triggered a problem.
We had a project which produced in frontend a critical error message, which was handled by WordPress  with his ""There has been a critical error on this website." output, delivered with a http-status-code 500.

But cachify saved the output and next time, nginx found the saved file and delivered it as expected with http-status-code 200.

So the problem is, that there shouldnt be a cache-file created if the http-status-code is not 200 in my opinion.
Perhaps you should add this check vor error-handling before you write the cache.

Thanks a lot

https://github.com/pluginkollektiv/cachify/blob/9ea110bfc39e677793884076b42b3a14f213da77/inc/class-cachify.php#L1395

karlkowald:
from default to statuscode matching 200

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1392,7 +1392,7 @@ final class Cachify {
 		 * @param string $cache_hash    The cache hash.
 		 * @param int    $cache_expires Cache validity period.
 		 */
-		$should_cache = apply_filters( 'cachify_store_item', true, $data, self::$method, self::_cache_hash(), self::_cache_expires() );
+		$should_cache = apply_filters( 'cachify_store_item', http_response_code() === 200, $data, self::$method, self::_cache_hash(), self::_cache_expires() );
 
 		/* Save? */
 		if ( $should_cache ) {
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1392,7 +1392,14 @@ final class Cachify {
 		 * @param string $cache_hash    The cache hash.
 		 * @param int    $cache_expires Cache validity period.
 		 */
-		$should_cache = apply_filters( 'cachify_store_item', http_response_code() === 200, $data, self::$method, self::_cache_hash(), self::_cache_expires() );
+		$should_cache = apply_filters(
+			'cachify_store_item',
+			200 === http_response_code(),
+			$data,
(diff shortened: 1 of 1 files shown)


inc/class-cachify.php
```diff
@@ -1392,7 +1392,7 @@ public static function set_cache( $data ) {
 		 * @param string $cache_hash    The cache hash.
 		 * @param int    $cache_expires Cache validity period.
 		 */
-		$should_cache = apply_filters( 'cachify_store_item', true, $data, self::$method, self::_cache_hash(), self::_cache_expires() );
+		$should_cache = apply_filters( 'cachify_store_item', http_response_code() === 200, $data, self::$method, self::_cache_hash(), self::_cache_expires() );
```
stklcode: [`http_response_code`](https://www.php.net/manual/function.http-response-code.php) was introduced in PHP 5.4

`Cachify` is yet compatible with 5.2

If we merge this for the next point or major release (likely 2.4), I'd vote for raising the minimum PHP version so something more reasonable above 5.4 and ignore this issue for now.

stklcode:
Besides the minor PHP version issue, code looks good to me. Thanks very much for this fix @karlkowald.

In general excluding 500s (which was the original intent) should be valid in almost any scenario.

Excluding errors like 404 is also a good idea in many cases. Caching them allows users/bots to fill up the cache requesting arbitrary, non-existing URLs. In the very same scene this wastes some processing time to not cache them.

I'm totally fine with a 200-only default here.

inc/class-cachify.php
stklcode: Seems we already have a vote for this in #259

So let's do it.

stklcode (approved):

inc/class-cachify.php
pfefferle: only one little thing: WordPress prefers yoda style checks

stklcode: If we're talking about style, we should add some line breaks, too.

pfefferle (approved):
LGTM!

PR #270, opened 2022-07-24
Change the CondPattern used to check if the request method is GET

angcl:
It was brought to Florian‘s and my attention that the CondPattern used does not work properly and the cache is done via the PHP fallback. Therefore, I hereby submit my proposal for a change.

Let‘s assume a page that is cached on the hard disk. The proposal then fulfils the following conditions:
- A GET request delivers the cache directly via the Apache HTTP server, PHP is not used.
- All other request methods return an uncached result.

I was able to test it at netcup and for me it works with the change as desired. Please check the behaviour again with different hosts.

### DIFF
--- a/inc/setup/cachify.hdd.htaccess.php
+++ b/inc/setup/cachify.hdd.htaccess.php
@@ -26,7 +26,7 @@ $beginning = '# BEGIN CACHIFY
   RewriteRule .* - [E=CACHIFY_DIR:/]
 {{GZIP}}
   # Main Rules
-  RewriteCond %{REQUEST_METHOD} ="GET"
+  RewriteCond %{REQUEST_METHOD} GET
   RewriteCond %{QUERY_STRING} =""
   RewriteCond %{REQUEST_URI} !^/(wp-admin|wp-content/cache)/.*
   RewriteCond %{HTTP_COOKIE} !(wp-postpass|wordpress_logged_in|comment_author)_


Zodiac1978:
This looks still wrong, looking at the docs
https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewritecond

> The string comparison operator is part of the CondPattern argument and must be included in the quotes if those are used. Eg.
RewriteCond %{HTTP_USER_AGENT} "=This Robot/1.0"

Or am I missing something here?

Zodiac1978:
Shouldn‘t review on mobile. This looks perfectly fine. Was just confused by the use of the „=„ here.

stklcode:
There are 8 working variants in Apache httpd 2.4:
```
RewriteCond %{REQUEST_METHOD} GET
RewriteCond %{REQUEST_METHOD} =GET
RewriteCond %{REQUEST_METHOD} "GET"
RewriteCond %{REQUEST_METHOD} "=GET"
RewriteCond "%{REQUEST_METHOD}" GET
RewriteCond "%{REQUEST_METHOD}" =GET
RewriteCond "%{REQUEST_METHOD}" "GET"
RewriteCond "%{REQUEST_METHOD}" "=GET"
```

For single-word patterns without brackets the quotes are not actually necessary. The quoted reference states that the operator (`=`) must be placed inside the quotes, if quotes are used. If there are none, there's no need to add them just because of the operator. It just says what we already discovered, i.e. `="GET"` is wrong.
(technically it is a valid pattern, but we are expecting `GET` here and don't want to match the two quotes)

The explicit equals sign is not strictly required either. Actually it's missing in almost any reference examples... I personally do prefer the short version that @angelocali used, for such simple one-word equality checks.

stklcode (approved):

PR #273, opened 2022-09-01
Fix content negotiation

### LINKED ISSUE #265: Cachify does not support Content Negotiation
**Describe the bug**
If a page has different views based on the Accept header, Cachify caches the output of the first request. If the request is `application/json` for example, Cachify always returns the JSON output on every following request.

**To Reproduce**
1. Install the ActivityPub plugin
2. Load the author page with Accept header `application/activity+json`
3. Load the author page again in the browser
4. See the JSON output

**Expected behavior**
Only cache requests with `html` Accept headers (https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values) and return the dynamic/uncached content otherwise.

Ancocodet:
Should fix this issue: https://github.com/pluginkollektiv/cachify/issues/265

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1258,6 +1258,13 @@ final class Cachify {
 			return true;
 		}
 
+		/* Content Negotiation */
+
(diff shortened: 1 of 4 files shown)


krafit:
LGTM

florianbrinkmann (approved):
LGTM

PR #276, opened 2022-11-18
fix URL parsing for cache key generation (#275)

### LINKED ISSUE #275: PHP warnings when generating cache key with latest 2.4 snapshot version
**Describe the bug**
This misbehavior is a regression from #190.

As a side effect the URL parsing routing has been modified:
https://github.com/pluginkollektiv/cachify/blob/d89efc317464d1d6df5a264786621e7a039bf465/inc/class-cachify.php#L1014-L1025

https://github.com/pluginkollektiv/cachify/blob/852cbd07aa431f37d68fad8bd01a67c94ec5ddfe/inc/class-cachify.php#L1000-L1011

If the method is called without arguments (which is done when the cache is populated), it raises a PHP warning:

> ( ! ) Warning: Undefined array key "host" in /var/www/html/wordpress/wp-content/plugins/cachify/inc/class-cachify.php on line 1081

We do set
```php
$url = wp_unslash( $_SERVER['HTTP_HOST'] ) . wp_unslash( $_SERVER['REQUEST_URI'] )
```
in this case, which is something like `wptest01.openshift.local/testpage/`. Passing this string to `wp_parse_url` results in a path-only array `[ "path" => "wptest01.openshift.local/testpage/`" ]`, so the `host` member is not available.

The resulting key is fine, because 
```php
'https-' . null . 'wptest01.openshift.local/testpage/'
```
and
```php
'https-' . 'wptest01.openshift.local' . '/testpage/'
```
are effectively the same. So caching actually works, but throws a warning.

**To Reproduce**
1. Install latest _Cachify_  development version on any compatible WordPress site
2. Enable the plugin
3. Enable PHP warnings in the log and/or output
4. Visit any yet uncached page

**Expected behavior**
The page should be displayed as usual and a cache entry should be generated in the corresponding backend (database, filesystem, ...) **without** any warnings.

**System:**
- OS: any
- WordPress: 6.1 (same for 6.0, 5.9 and likely any compatible version)
- PHP: 8.1 (same for 8.0 and 7.4)
- Version: https://github.com/pluginkollektiv/cachify/commit/09bdf189e7c85701fa7f4b817bdbb8043f8c1ba2

**Additional context**
Enable PHP warnings in log and/or output, e.g. setting `define('WP_DEBUG', true);`

stklcode:
As part of refactoring in #190 the URL parsing has been altered, s.t. we pass a concatenation of host and path to `wp_parse_url()`. This results in the actual host not being parsed at all which raises a PHP warning.

Prefix the URL with double slashes (the protocol is not of interest here) to fix the parsing.

We could use a full URL here, like
```php
$url = wp_unslash( $_SERVER['REQUEST_SCHEME'] ) '://' . wp_unslash( $_SERVER['HTTP_HOST'] ) . wp_unslash( $_SERVER['REQUEST_URI'] );
```
or even add port, query, etc. to make it complete, but we only process host and path in the very next lines, so a reduced, protocol-agnostic URL is sufficient here.

resolves #275

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1074,7 +1074,7 @@ final class Cachify {
 
 		if ( empty( $url ) ) {
 			// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.InputNotValidated
-			$url = wp_unslash( $_SERVER['HTTP_HOST'] ) . wp_unslash( $_SERVER['REQUEST_URI'] );
+			$url = '//' . wp_unslash( $_SERVER['HTTP_HOST'] ) . wp_unslash( $_SERVER['REQUEST_URI'] );
 		}
 
 		$url_parts = wp_parse_url( $url );


florianbrinkmann (approved):
LGTM

PR #278, opened 2022-11-20
move dark-mode styling into dashboard.css

### LINKED ISSUE #264: Styling for dark mode plugin that is a Markdown editor now
A few years ago a style modification for the dashboard icon was added to look good when used with a dark mode plugin. That plugin contains a markdown editor now and does not call the filter that was used back then. Should we just remove the style modification or keep it for users that stick with the old dark mode plugin version?

We (@angelocali and I) also thought about adding support for the latest version of the plugin, because besides the markdown editor option (that is a paid feature only), the dark mode is still there. But then we decided to not add it, because rebranding a plugin from dark mode to markdown editor is a little bit creepy and nothing we wanted to add support for.

stklcode:
Dark mode styling has been added inline using the "doing_dark_mode" hook which is no longer called in current plugin versions.

Include the CSS definition directly into dashboard.css and deprecate the hook for adding inline styles.

----

resolves #264 

**However:** This does _not_ apply to all "dark-mode" plugins and not even to all themes, because the icon has two different shades of gray and is not affected by most stylings that assume Dashicons are used here. We might prefer a different approach like #277

### DIFF
--- a/css/dashboard.css
+++ b/css/dashboard.css
@@ -5,6 +5,10 @@
 	fill: #606a73;
 }
 
+html.dark-mode body:not(.block-editor-page) #dashboard_right_now .cachify-icon {
+	fill: #bbc8d4;
+}
+
 #dashboard_right_now li a.cachify-glance::before {
 	content: "";
 	padding: 0;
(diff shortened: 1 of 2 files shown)


pfefferle (approved):
LGTM

PR #279, opened 2022-11-20
clean up PHPdoc blocks

stklcode:
Changes apply to comment blocks only, no code modified.

* consistent use of `int`/`bool` over `integer`/`boolean`
* consistent indentation of `@param` and `@return` blocks
* use `@since` tags instead of `@change`, where applicable
  `@change` is no valid PHPdoc tag and is has often been used without any context information. Use multiple `@since` declarations for external changes (visibility, arguments, renaming, ...) and remove behavior-related `@change` tags.
* remove redundant `@change` tags
  there is no additional information whatsoever in blocks like
```php
  * @since   2.0.7
  * @change  2.0.7
```
* use version number that actually exist (2.0, 2.3.0, ...)
* remote superfluous `true/false`  additions from tags like `@return bool true/false <description>`

### DIFF
--- a/inc/class-cachify-apc.php
+++ b/inc/class-cachify-apc.php
@@ -16,10 +16,9 @@ final class Cachify_APC {
 	/**
 	 * Availability check
 	 *
-	 * @since   2.0.7
-	 * @change  2.0.7
+	 * @return bool TRUE when installed
 	 *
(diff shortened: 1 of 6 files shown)


pfefferle (approved):
LGTM

PR #286, opened 2023-08-07
invalidate cache when permalink changes

### LINKED ISSUE #285: Cache doesn't invalidate when permalink changes
**Describe the bug**

The cache doesn't invalidate when the permalink of a page/post changes.

I use "HDD" mode (which is a bit misleading, because files are stored on SSD) with the proposed changes to `.htaccess`.

**To Reproduce**

* create page with permalink `/test/`
* Open `/test` in browser to trigger caching of static file in `/wp-content/cache/cachify/{host}/test/index{.html,.html.gz}`
* Change permalink to `/test2/`
* `/test/` folder with html file still exists in cache folder

**Expected behavior**

* When a permalink changes, the cached files `/test/index.html` and `/test/index.html.gz` should be deleted.
* Also the parent folder `/test/` should be deleted.

**System:**

* OS: Local test instance, using docker with a modified version of `wordpress:php8.1-apache` (e. g. `a2enmod rewrite expires headers`)
* WordPress: 6.3-RC3-56351
* Cachify: 2.3.2

**Solution**

Use `post_updated` action instead of `save_post` and check against the original post to invalidate the cache with the original permalink.

* https://github.com/pluginkollektiv/cachify/blob/develop/inc/class-cachify.php#L96
* https://github.com/pluginkollektiv/cachify/blob/develop/inc/class-cachify.php#L1009

While there, internal methods should pass `WP_Post` instead of `int $post_id` to save a few db calls.

For reference:

* https://developer.wordpress.org/reference/hooks/post_updated/
* https://developer.wordpress.org/reference/hooks/save_post/
* https://developer.wordpress.org/reference/functions/get_permalink/

I could send a pull request with changes, but I'm not sure, if this plugin is still actively developed.

**Other**

I do some research about WP Cache plugins and Cachify looked promising (no bloat, readable code and a rant about CDNs in the docs).

First I tested WP Super Cache, but it loads the whole settings ui on each frontend call, which slows each non-cached request down. Also the code is unreadable.

Cachify is number 2 on my list. I like it, but as mentioned above, I would like to use (and contribute to) a maintained plugin.

Cache Enabler would be the next candidate. I read in an issue here, that it is a fork of Cachify. <del>At first I wanted to drop that because of unhealthy permissions, but I saw, that Cachify has the same issue (`chmod` with 777 and 666)</del> *edit:* While skimming over the code I misunderstood the meaning of `$stat['mode'] & 0007777`.

raffaelj:
Fixes #285

### DIFF
--- a/inc/class-cachify-hdd.php
+++ b/inc/class-cachify-hdd.php
@@ -229,7 +229,7 @@ final class Cachify_HDD {
 	 * @since 2.0
 	 */
 	private static function _clear_dir( $dir, $recursive = false ) {
-		/* Remote training slash */
+		/* Remove trailing slash */
 		$dir = untrailingslashit( $dir );
 
 		/* Is directory? */
--- a/inc/class-cachify-hdd.php
+++ b/inc/class-cachify-hdd.php
@@ -99,8 +99,7 @@ final class Cachify_HDD {
 	 */
 	public static function delete_item( $hash, $url ) {
 		self::_clear_dir(
-			self::_file_path( $url ),
-			true
+			self::_file_path( $url )
 		);
 	}
 
--- a/inc/class-cachify-hdd.php
+++ b/inc/class-cachify-hdd.php
@@ -99,7 +99,8 @@ final class Cachify_HDD {
 	 */
 	public static function delete_item( $hash, $url ) {
 		self::_clear_dir(
-			self::_file_path( $url )
+			self::_file_path( $url ),
+			true
 		);
 	}
 
--- a/inc/class-cachify-hdd.php
+++ b/inc/class-cachify-hdd.php
@@ -228,7 +228,7 @@ final class Cachify_HDD {
 	 * @since 2.0
 	 */
 	private static function _clear_dir( $dir, $recursive = false ) {
-		/* Remove trailing slash */
+		/* Remote training slash */
 		$dir = untrailingslashit( $dir );
 
 		/* Is directory? */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file. This projec
 
 Requires PHP 5.6 and WordPress 4.7 or above
 
+* Fix: invalidate cache when permalink changes (#285, #286, props raffaelj)
 * Enhance: adjust styling for setup instructions (#215, props timse201)
 * Enhance: update hooks for Multisite initialization in WordPress 5.1 and above (#246, props ouun)
 * Enhance: rework flush hooks and add some third-party triggers for Autoptimize and WooCommerce (#225, props timse201)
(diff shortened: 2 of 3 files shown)


inc/class-cachify-hdd.php
```diff
@@ -99,7 +99,8 @@ public static function get_item() {
 	 */
 	public static function delete_item( $hash, $url ) {
 		self::_clear_dir(
-			self::_file_path( $url )
+			self::_file_path( $url ),
+			true
```
stklcode: I wonder if this is really the best way so solve your problem here.

I assume you added the `$recursive = true`, because in case you change the the permalink if `/foo` to `/bar` then also subpages like `/foo/sub` move to `/bar/sub`.

But if you just change some content of `/foo`, why should all subpages always be pruned, too? Imagine a site with only a few top-level paged but a huge subpage structure, this might introduce a lot of avoidable overhead.

raffaelj: When I added `$recursive = true`, I didn't really think about cached sub pages. My main concern was to delete the empty folder after invalidating the page cache.

These orphaned, empty folders aren't harmful, so I'll revert this change.

I think, it would be better to add an additional check, if the folder is empty. If so, the empty folder should be deleted by default - regardless of the `$recursive = false` option.

stklcode: For the empty folder I would simply remove the `if ($recursive )` condition before checking if the directory is empty. Deleting an empty directory in non-recursive is okay, no idea why there second condition exists.

raffaelj: > For the empty folder I would simply remove the `if ($recursive )` condition

Sounds good.

I would also delete the early return, which has no benefit, but prevents deleting empty folders:

https://github.com/pluginkollektiv/cachify/blob/develop/inc/class-cachify-hdd.php#L245-L248

```php
		/* Empty? */
		if ( empty( $objects ) ) {
			return;
		}
```

stklcode: Yes, absolutely.
But I would rather do this in a separate change and merge this PR as is.

stklcode (approved):

stklcode:
Thanks very much @raffaelj for your work from analysis to this fix and also the input on the empty dir issue.

Let's see if we can get the version done and ship the fix as soon as possible.

inc/class-cachify-hdd.php
stklcode: Did so in #289 with some additional refactoring and unit tests. Didn't want to bloat your PR with all this stuff.

raffaelj:
> But I would rather do this in a separate change and merge this PR as is.
> Did so in https://github.com/pluginkollektiv/cachify/pull/289 with some additional refactoring and unit tests. Didn't want to bloat your PR with all this stuff.

Thanks.

> Let's see if we can get the version done and ship the fix as soon as possible.

That would be great.

PR #290, opened 2023-08-22
Make change to robots.txt configurable

### LINKED ISSUE #288: Make change to robots.txt configurable
As described in #282, `robots.txt` produces invalid output. I don't need that change to robots.txt at all, because I deny direct access to cached files via `.htaccess`:

```
# Deny direct access to cache folder, but allow internal redirect
RewriteCond %{ENV:REDIRECT_STATUS} ^$
RewriteRule ^wp-content/cache/cachify/ - [F]
```

**Expected behavior**

I want to click a checkbox in the settings ui to disable automatic changes to `robots.txt`. When disabled, the hook `do_robots` is skipped.

**System:**

* test docker instance with apache, php81
* WP 6.3
* Cachify dev branch, hdd mode

raffaelj:
Fixes #288


inc/class-cachify.php
```diff
@@ -422,15 +422,16 @@ private static function _get_options() {
...
+				'only_guests'       => 1,
+				'compress_html'     => self::MINIFY_DISABLED,
+				'cache_expires'     => 12,
+				'without_ids'       => '',
+				'without_agents'    => '',
+				'use_apc'           => self::METHOD_DB,
+				'reset_on_post'     => 1,
+				'reset_on_comment'  => 0,
+				'sig_detail'        => 0,
+				'change_robots_txt' => 0,
```
stklcode: I suggest enabling this feature by default, so we do not silently disable it when upgrading existing installations.

```suggestion
				'change_robots_txt' => 1,
```

stklcode (changes requested):

inc/class-cachify.php
raffaelj: I disabled it by default because the last state caused a broken setup. Now with #283 merged it should be fine to change it back.

stklcode (approved):

stklcode:
Thanks Raffael, merged the feature.

PR #294, opened 2023-11-03
Adds action triggers to cache flush methods - Allows Cachify flushing to be extended

ouun:
I tried to extend Cachify with hooks to handle additional API requests to an edge cache. However, I ended up with some very hacky solutions as there are no action hooks available at the end of the cache flush/delete logics.

This PR simply adds two action triggers to the two main cache flush methods, which I hope I have identified correctly. It allows for any other callback functions to be added after the entire cache or a single cache item has been deleted.

**I really hope** you agree with this and that **it can be merged with the 2.4.0 release**. Redis support and these hooks will make Cachify even more flexible and powerful. Thanks for your great work!

### DIFF
--- a/inc/class-cachify.php
+++ b/inc/class-cachify.php
@@ -1170,16 +1170,16 @@ final class Cachify {
 		}
 
 		$hash = self::_cache_hash( $url );
-		call_user_func(
-			array(
-				self::$method,
-				'delete_item',
-			),
-			$hash,
-			$url
-		);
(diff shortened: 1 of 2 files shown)


stklcode:
While being pretty much self-explaining and matching calling method's names, I find the hook names little long.

Maybe something like `cachify_flushed_page` and `cachify_flushed_total`? Or simply keep the long names, I only have a weak personal opinion on this one...

stklcode (approved):

PR #295, opened 2023-11-11
fix parameter use of `trigger_error(esc_html(...))`

stklcode:
The second parameter was supposed to be passed to `trigger_error`, not `esc_html`. Move the closing parenthesis to resolve this issue.


pfefferle (approved):
LGTM

PR #296, opened 2023-11-11
update WPCS to v3 and apply minor code style tweaks

stklcode:
* use `__DIR__` instead of `dirname( __FILE__ )`
* remove parentheses from `require_once` calls
* rename `$class` parameter of autoload function to `$class_name`
* fix precision alignmeht, equals signs and double arrows
* update WPCS definitions to v3.0


pfefferle (approved):
LGTM

PR #298, opened 2024-04-06
Common interface for the Cachify storage classes

lloc:
I started to collect the least common methods with their arguments. It is now possible to guarantee a common interface by implementing **Cachify_Backend** in the interested classes.

I also made the overall coverage more visible and narrowed the output down to the classes (and not template-like the setup code for example):

image

### DIFF
--- a/cachify.php
+++ b/cachify.php
@@ -91,7 +91,7 @@ spl_autoload_register( 'cachify_autoload' );
  * @param string $class_name the class name.
  */
 function cachify_autoload( $class_name ) {
-	if ( in_array( $class_name, array( 'Cachify', 'Cachify_APC', 'Cachify_DB', 'Cachify_HDD', 'Cachify_MEMCACHED', 'Cachify_REDIS', 'Cachify_CLI' ), true ) ) {
+	if ( in_array( $class_name, array( 'Cachify', 'Cachify_APC', 'Cachify_Backend', 'Cachify_CLI', 'Cachify_DB', 'Cachify_HDD', 'Cachify_MEMCACHED', 'Cachify_REDIS' ), true ) ) {
 		require_once sprintf(
 			'%s/inc/class-%s.php',
 			CACHIFY_DIR,
--- a/composer.json
+++ b/composer.json
@@ -68,6 +68,9 @@
     ],
     "test": [
       "phpunit"
+    ],
+    "coverage": [
+      "php -d xdebug.mode=coverage vendor/bin/phpunit --coverage-html tests/coverage"
     ]
   },
   "config": {
--- a/inc/class-cachify-memcached.php
+++ b/inc/class-cachify-memcached.php
@@ -11,7 +11,7 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Cachify_MEMCACHED
  */
-final class Cachify_MEMCACHED {
+final class Cachify_MEMCACHED implements Cachify_Backend {
 
 	/**
 	 * Memcached-Object
@@ -140,9 +140,12 @@ final class Cachify_MEMCACHED {
 	/**
 	 * Print the cache
 	 *
+	 * @param bool  $sig_detail Show details in signature.
+	 * @param array $cache      Array of cache values.
+	 *
 	 * @since 2.0.7
 	 */
-	public static function print_cache() {
+	public static function print_cache( $sig_detail, $cache ) {
 		// Not supported.
 	}
 
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules/
 composer.lock
 package-lock.json
 phpunit.*.xml
+tests/coverage/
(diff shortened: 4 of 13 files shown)


inc/class-cachify-interface.php
```diff
@@ -0,0 +1,68 @@
...
+ * @package Cachify
+ */
+
+/* Quit */
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Interface_Cachify_Storage
+ */
+interface Cachify_Interface {
```
stklcode: The `Cachify_` prefix is used like a namespace substitute, so the actual interface name is `Interface`. Maybe something like `Cachify_Backend` or similar would be little more descriptive.

inc/class-cachify-interface.php
```diff
@@ -0,0 +1,68 @@
+<?php
+/**
+ * Interface for APC based caching storage classes.
```
stklcode:
```suggestion
 * Interface for caching storage classes.
```

Well, one implementation is APC, but the others are not. The option name is just a relict of the early days where APC was the only alternative backend.

inc/class-cachify-interface.php
```diff
@@ -0,0 +1,68 @@
...
+	 * @return bool TRUE if available
+	 */
+	public static function is_available();
+
+	/**
+	 * Caching method as string
+	 *
+	 * @return  string Caching method
+	 */
+	public static function stringify_method();
```
stklcode: I do dislike this method name and would have called it something as simple as `get_name()`. But that should probably be done separately as this is only a copy of the pre-existing implementation.

inc/class-cachify-memcached.php
```diff
@@ -140,9 +140,12 @@ public static function clear_cache() {
 	/**
 	 * Print the cache
 	 *
-	 * @since 2.0.7
+	 * @param bool  $sig_detail Show details in signature.
+	 * @param array $cache      Array of cache values.
+	 *
+	 * @since 2.0
```
stklcode: Why was the `.7` dropped?
Not that anybody really cares, but up to 2.0.4 the class did not even exist, so 2.0 is strictly speaking not correct. There is no history between 2.0.4 and 2.1.6 available. In a SemVer world 2.1 would make sense, but I’d guess 2.0.7 is plausible, too.

cachify.php
```diff
@@ -91,7 +91,7 @@
  * @param string $class_name the class name.
  */
 function cachify_autoload( $class_name ) {
-	if ( in_array( $class_name, array( 'Cachify', 'Cachify_APC', 'Cachify_DB', 'Cachify_HDD', 'Cachify_MEMCACHED', 'Cachify_REDIS', 'Cachify_CLI' ), true ) ) {
+	if ( in_array( $class_name, array( 'Cachify', 'Cachify_APC', 'Cachify_DB', 'Cachify_HDD', 'Cachify_MEMCACHED', 'Cachify_REDIS', 'Cachify_CLI', 'Cachify_Interface' ), true ) ) {
```
stklcode: I would prefer alphabetical order here, although `CLI` is any outlier already.

composer.json
```diff
@@ -68,6 +68,9 @@
     ],
     "test": [
       "phpunit"
+    ],
+    "coverage": [
+      "php -d xdebug.mode=coverage vendor/bin/phpunit --coverage-html tests/coverage"
```
stklcode, line 73: Should we migrate CI pipelines to actually track coverage (e.g. run Sonarqube analysis explicitly after test execution) or would that be too frustrating at the moment? 🙈

stklcode (commented):
Good work, thanks. I would have some minor cosmetic/pedantic suggestions, but I totally agree with the changes proposed.

composer.json
lloc: SonarCloud/SonarCube would offer Quality gates for coverage on new code which I like very much. I see this here as "on top". Just to get it quickly in your local development.

inc/class-cachify-interface.php
lloc: Yeah, I adjusted it by iteration. After realizing there is an autoloader with a naming pattern. ;-) But yours is good input, I will go to change it.

inc/class-cachify-interface.php
lloc: Yes I guess we can refactor the name after this PR.

inc/class-cachify-memcached.php
lloc: Good point. I missed the knowledge of the past development here (plus copy & past :see_no_evil: ).

cachify.php
lloc: Missed totally, that this was in alphabetical order. Thanks!

lloc:
> Good work, thanks. I would have some minor cosmetic/pedantic suggestions, but I totally agree with the changes proposed.

Thanks for taking the time to review the PR!

composer.json
stklcode: "New code" can be pretty much everything when refactoring, reformatting or introducing new hierarchies though, especially on small projects.

We should enable analysis in a follow-up change to not block this one here.

stklcode (approved):
Squashed the changes into 2 commits (interface + coverage), LGTM

PR #304, opened 2024-08-06
remove APC support

### LINKED ISSUE #305: Show a notice, if the configured caching method is no longer available
We currently check for availability of the backend, e.g. if if a compatible PHP extension is available for external object caches or permalink structure is active for HDD caching. If this is not the case, we silently fall back to Database Caching which is always available.

We should at least add an admin notification to attract attention, because the plugin configuration does not fit the server environment which leads to potentially unintended behavior.

stklcode:
APC is only supported up to PHP 5.5. As Cachify now requires PHP 5.6 or later, there is no point in maintaining legacy APC support.

We also introduce an admin notice, if the configured backend is not available. This applies to all backends except DB.

cachify_notice_apc


tests/test-cachify-noop.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * Cachify HDD tests.
```
florianbrinkmann:
```suggestion
 * Cachify NOOP tests.
```

florianbrinkmann (approved):

PR #306, opened 2024-09-22
docs: prune tags in readme to 5

stklcode:
**keep:** cache, caching, performance, speed

**remove:** acceleration, minimize, apc, disk, hdd, memcached, compression, minify

**add:** optimize

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,7 +1,7 @@
 # Cachify #
 * Contributors:      pluginkollektiv
 * Donate link:       https://www.paypal.com/cgi-bin/webscr?cmd=_donations&business=TD4AMD2D8EMZW
-* Tags:              acceleration, cache, caching, minimize, performance, apc, disk, hdd, memcached, compression, minify, speed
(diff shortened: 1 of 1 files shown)


lloc:
The set of keywords seems ok to me. Other plugins of the same category use `core web vitals` and similar. Introducing `speed` may be fine too.

stklcode:
Plugins with the proposed 5 tags:

* https://wordpress.org/plugins/tags/performance/ (387)
* https://wordpress.org/plugins/tags/cache/ (290)
* https://wordpress.org/plugins/tags/speed/ (177)
* https://wordpress.org/plugins/tags/optimize/ (133)
* https://wordpress.org/plugins/tags/caching/ (98)

and

* https://wordpress.org/plugins/tags/core-web-vitals/ (29)

I do not have a strong opinion on this one. `cache` + `caching` seems redundant, `performance` and `optimize` little too general... Just removed some, but open to replace any of them.

stklcode:
Let's go with the proposed 5. We can simply update the readme again if we want to optimize the tags again.

PR #308, opened 2024-10-06
fix: test if gzip extension is available

stklcode:
If _zlib_ PHP extension is not loaded and there is no substitute in place, `gzencode()` does not exist and generation of compressed files will fail.

Add a simple `function_exists()` test to prevent this.

### DIFF
--- a/inc/class-cachify-hdd.php
+++ b/inc/class-cachify-hdd.php
@@ -33,6 +33,11 @@ final class Cachify_HDD implements Cachify_Backend {
 	 * @since 2.4.0
 	 */
 	public static function is_gzip_enabled() {
+		if ( ! function_exists( 'gzencode' ) ) {
+			// GZip is not available on the system.
+			return false;
+		}
+
 		/**
 		 * Filter that allows to enable/disable gzip file creation
 		 *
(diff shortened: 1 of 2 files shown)


stklcode:
It might seem little far fetched, but providing the `function_exists` result as default filter parameter allows to implement some logic to define this function on demand, e.g.

```php
add_filter( 'cachify_create_gzip_files', function ( $available ) {
  if ( ! $available ) {
    // Define or load anything that implements gzencode.
    function gzencode( $data, $level ) {
      return $data;
    }
  }

  return true;
} );
```

Zodiac1978 (commented):
Looks good to me!

Just one thing: it would still be possible to use the filter although gzencode is not available, or not? I know this would be the problem of the one using the filter, but we could also safeguard this. What do you think?

stklcode:
In the current suggestion: Yes, you can still override the `false` with a filter, if `gzencode` is not available.

One might polyfill another compression routine on demand in the filter hook (see example above). Seems more of a theoretical possibility though.

But I'd be fine changing the logic again, so the filter is completely omitted if `gzencode` is not available. No hard preference for either way.

stklcode:
I changed the logic, so if `gzencode` is not available we disable the feature and do not execute the filter.

This is trivial to change again, but >99.9% should not need such tricky configurations.

PR #309, opened 2024-10-06
fix: remove trailing semicolon from .htaccess sample code

stklcode:
Fix regression from #302 where `&gt` was replaced with `>` instead of `&gt;` which makes the apache config syntactically invalid.

```
# BEGIN CACHIFY
<IfModule mod_rewrite.c>;
                        ^
```

Removing the semicolons should make the config valid again.


Zodiac1978 (approved):
LGTM!

PR #317, opened 2025-02-01
fix validation of redis context parameters array (#315)

### LINKED ISSUE #315: Valdation of Redis servers rejects auth parameter
**Describe the bug**

Redis server configuration rejects valid auth parameter.

**To Reproduce**

On a system with Redis support, add a filter for Redis servers with 7th (auth) parameter:

```php
add_filter( 'cachify_redis_servers', fn () => [
    '127.0.0.1', 6379, 1, null, 0, 0,
    [ 'auth' => [ 'username', 'password' ] ]
] );
```

Validation rejects the last parameter, because the condition is invalid:

Internally: `Cachify_REDIS::sanitize_con_parameters()` returns `false` because of

```php
if ( count( $con ) > 6 && ! is_null( $con[6] ) && is_array( $con[6] ) ) {
//                                             ^^^^^^^^
	return false;  // Context parameters, e.g. authentication (since PhpRedis 5.3).
}
```

**Expected behavior**

Cachify should try to connect to the configured Redis server using "username:password" credentals.

I.e. `Cachify_REDIS::sanitize_con_parameters()` should return the (sanitized) array.

**System:**
- Cachify 2.4.0

**Additional context**

Cross reference: https://wordpress.org/support/topic/php-fatal-error-457

stklcode:
The optional 7th argument the Redis::connect() holds an array of context parameters, required for authentication. This parameter can either be NULL or an array.

Fix the condition, so it does not reject arrays.

resolves #315

### DIFF
--- a/inc/class-cachify-redis.php
+++ b/inc/class-cachify-redis.php
@@ -271,7 +271,7 @@ final class Cachify_REDIS implements Cachify_Backend {
 			if ( count( $con ) > 5 ) {
 				$con[5] = floatval( $con[5] );  // Read timeout in seconds.
 			}
-			if ( count( $con ) > 6 && ! is_null( $con[6] ) && is_array( $con[6] ) ) {
+			if ( count( $con ) > 6 && ! is_null( $con[6] ) && ! is_array( $con[6] ) ) {
 				return false;  // Context parameters, e.g. authentication (since PhpRedis 5.3).
 			}
 			if ( count( $con ) > 7 ) {


lloc (commented):
That seems OK to me. I guess this becomes more and more complex. I will do a PR and bring PHPStan in. ;-)

stklcode:
Yes, this validation is not very pretty.

And spoiler alert: it becomes even more of a mess if we want to extend it to accept PHPRedis 6 syntax

```php
[
    'host' => '127.0.0.1',
    'port' => 6379,
    'connectTimeout' => 2.5,
    'auth' => ['phpredis', 'phpredis'],
    'ssl' => ['verify_peer' => false],
    'backoff' => [
        'algorithm' => Redis::BACKOFF_ALGORITHM_DECORRELATED_JITTER,
        'base' => 500,
        'cap' => 750,
    ],
]
```

We should probably think about dropping the deep sanitization, advanced Redis config is a "pro-feature" anyway.

PR #320, opened 2025-02-11
Introduce PHPStan and adjust minimum requirements  

lloc:
This PR introduces **PHPStan** with the initial level set to **1**, with the goal of gradually increasing it to at least **5** or higher over time.  

### Changes and Improvements  

- **PHPStan Integration**  
  - PHPStan can be executed locally using `composer phpstan`.  
  - It has also been added as an additional **GitHub workflow** for automated checks.  

- **Minimum PHP and WordPress Requirements**  
  - PHPStan requires at least **PHP 7.2**, so this PR suggests:  
    - Dropping support for **PHP 5.6**.  
    - Setting the minimum requirements to **WordPress 5.1** and **PHP 7.2**.  
  - As a sidenote, WordPress 5.1 includes a fix for a bug in the test suite introduced after version 4.x.  

- **PHP 8.4 Compatibility**  
  - **PHP 8.4** has been added to the test checks.

### DIFF
--- a/.github/workflows/phpstan-ci.yml
+++ b/.github/workflows/phpstan-ci.yml
@@ -1,0 +1,21 @@
+name: Run PHPStan
+
+on: [push, pull_request]
+
+jobs:
+  run-phpstan:
+    name: Static Analysis
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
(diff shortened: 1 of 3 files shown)


stklcode (approved):
Rebased on top if #323 where we raised the requirements to PHP 7.2 and WP 5.1, so these changes became obsolete here.
Also added `pull_request` trigger to the new CI workflow.

Analysis now runs and brings up 2 findings which are valid and should be addressed in another PR.

PR #321, opened 2025-02-27
move select helpers from private methods into settings file

stklcode:
Both helpers are only used exactly once when building the settings view. Move the logic into the settings file itself, initialize arrays directly and remove double HTML escaping.


lloc:
This is ok for me!

PR #324, opened 2025-03-08
Use prefixed Redis keys to prevent data loss on cache clearing.

sdellenb:
Only cachify entries that belong to the current host are removed from the cache.

### DIFF
--- a/inc/class-cachify-redis.php
+++ b/inc/class-cachify-redis.php
@@ -116,6 +116,7 @@ final class Cachify_REDIS implements Cachify_Backend {
 		}
 
 		/* Delete all cache entries for this site */
+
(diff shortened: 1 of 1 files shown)


inc/class-cachify-redis.php
```diff
@@ -115,8 +115,21 @@ public static function clear_cache() {
 			return;
 		}
 
-		/* Flush */
-		@self::$_redis->flushAll();
+		/* Delete all cache entries for this site */
+		$host = wp_unslash( $_SERVER['HTTP_HOST'] );
+		$keys = self::$_redis->keys( $host . '*' );
```
sdellenb: `keys()` returns the keys **_with_** the configured prefix, but `del()` or `unlink()` do not expect it and will add it again before sending the command to Redis.

This quirk of phpredis was discussed in https://github.com/phpredis/phpredis/issues/1117#issuecomment-283159870 and only has to be considered when bulk-deleting.

stklcode (approved):

PR #326, opened 2025-04-16
merge hooks for flush icon style and script and use plugin version instead of filemtime

### LINKED ISSUE #312: Missing dependencies shown in Query Monitor
Query Monitor is showing missing dependencies for styles and scripts with the handle `cachify-admin-bar-flush`:
Bildschirmfoto 2024-11-13 um 16 49 04

The link to the source is showing the correct file content.

My install is in a subfolder. Maybe this is a problem here? Or sharing the same handle for style and script is?

stklcode:
Fixes and sideshows of #312.

We have a `admin_bar_menu` hook to add the "flush cache" icon including stylesheet and another hook in the same phase to add the script.
We also determine file modification times of all 3 resources using `filemtime()` on the CSS/JS file.
And finally, a possible race condition inside the Javascript, if the icon is not rendered before script excution.

1. Append constant `CACHIFY_VERSION` instead of dynamic timestamps.
2. Merge script and style hooks and move them to `wp_enqueue_scripts`/`admin_qneueue_scripts` phase.
3. Rework JS to use `DOMContentLoaded´ event listener.

### DIFF
--- a/cachify.php
+++ b/cachify.php
@@ -41,6 +41,7 @@ define( 'CACHIFY_FILE', __FILE__ );
 define( 'CACHIFY_DIR', __DIR__ );
 define( 'CACHIFY_BASE', plugin_basename( __FILE__ ) );
 define( 'CACHIFY_CACHE_DIR', WP_CONTENT_DIR . '/cache/cachify' );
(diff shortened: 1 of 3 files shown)


Zodiac1978 (approved):
LGTM!

PR #329, opened 2025-07-31
refactor: introduce a static void function init() instead of constructor

stklcode:
Calling `instance() `constructs an unused `Cachify` instance. Extract the (static) logic from the constructor into a new void methd `init()` and use it as initialization hook.

### DIFF
--- a/cachify.php
+++ b/cachify.php
@@ -49,7 +49,7 @@ add_action(
 	'plugins_loaded',
 	array(
 		'Cachify',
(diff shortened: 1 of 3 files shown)


2ndkauboy:
Looks good and makes sense!

stklcode:
PHPStan is still failing, but with only 1 finding instead of 2, so let's see it as partial success :wink:

PR #339, opened 2026-07-05
Add blueprint file to trigger preview feature

Drivingralle:
Add the blueprint file inside the required folder to make the preview on the plugin page in WP.org repo work

### DIFF
--- a/.wordpress-org/blueprints/blueprint.json
+++ b/.wordpress-org/blueprints/blueprint.json
@@ -1,0 +1,30 @@
+{
+	"$schema": "https://playground.wordpress.net/blueprint-schema.json",
+	"preferredVersions": {
+		"php": "8.4",
+		"wp": "latest"
+	},
+	"phpExtensionBundles": [
+		"kitchen-sink"
+	],
+	"features": {
+		"networking": true
+	},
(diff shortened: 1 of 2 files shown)


.wordpress-org/blueprints/blueprint.json
```diff
@@ -0,0 +1,30 @@
...
+	},
+	"landingPage": "/wp-admin/options-general.php?page=cachify",
+	"login": {
+		"username": "admin",
+		"password": "password"
+	},
+	"steps": [
+		{
+			"step": "installPlugin",
+			"pluginZipFile": {
```
stklcode: I know this used to work fine, but from the docs shouldn’t pluginData be the preferred version?
https://wordpress.github.io/wordpress-playground/blueprints/data-format#use-one-plugin-install-shape

```suggestion
			"pluginData": {
```

Also seen the shorthand I wasn’t aware of:
```json
  "plugins": [ "cachify" ]
```

We have a mix of with/without installPlugin across our projects. For w.org preview this is not necessary, but for portability it doesn’t hurt go keep it here.

Drivingralle: I like the shorthand. Didn't know about that one, too.

stklcode (approved):
LGTM

